PR #8, opened 2020-02-02
No longer pin numpy for pylint after pylint upgrade, for Azure and Gi…

inducer:
…tlab


inducer:
https://gitlab.tiker.net/inducer/pytential/merge_requests/190

PR #21, opened 2020-07-18
Re-revert `test-matrix-refactor` branch

alexfikl:
This re-reverts the branch after [this](https://gitlab.tiker.net/inducer/pytential/-/merge_requests/210). 

The failing tests now give exactly the same results as before. Fingers crossed!


inducer:
:shrug: So the automatic refinement ran it out of memory? LGTM. Thanks!

alexfikl:
> So the automatic refinement ran it out of memory? LGTM. Thanks!

Yeah, the torus was too thin and got refined to about 10k nodes, while before it was just 3k nodes. These tests build the full dense matrix, so they easily use a lot of memory.

Issue #27, opened 2020-09-08
from __future__ import annotations is python 3.7+


isuruf:
Is this used anywhere? If not, we can remove it to make pytential 3.6 compatible.

inducer:
I'm not sure those are essential. Could you submit a PR that removes them to see what happens?

PR #28, opened 2020-09-08
Remove importing annotations

isuruf:
Makes pytential compatible with python 3.6

Fixes https://github.com/inducer/pytential/issues/27


inducer:
Thanks!

PR #31, opened 2020-09-25
Allow point nodes in map_node_coordinate_component

alexfikl:


pytential/symbolic/execution.py
```diff
@@ -204,7 +204,16 @@ def map_ones(self, expr):
...
                 expr.dofdesc.geometry, expr.dofdesc.discr_stage)
-        return thaw(self.array_context, discr.nodes()[expr.ambient_axis])
+
+        from pytential.target import PointsTarget
+        from pytential.source import PointPotentialSource
+        from meshmode.discretization import Discretization
+        if isinstance(discr, Discretization):
+            return thaw(self.array_context, discr.nodes()[expr.ambient_axis])
+        elif isinstance(discr, (PointPotentialSource, PointsTarget)):
+            return discr.nodes()[expr.ambient_axis]
```
inducer, line 214: Why aren't they thaw-able?

alexfikl: They're not thaw-able with that `thaw` because they're not necessary `DOFArray`s. Maybe cleaner way would be
```python
x = discr.nodes()[expr.ambient_dim]
if isinstance(x, DOFArray):
    return thaw(...)
else:
   return actx.thaw(...)
```
?

inducer: That sounds good to me. Thanks for suggesting it.

inducer:
Thanks!

PR #34, opened 2020-10-19
Be more strict about input to IntG.

isuruf:
There's really no reason to allow IntG(0) for Laplace
and IntG("k") for Helmholtz


inducer:
Fair enough, thanks!

PR #43, opened 2020-11-27
Handle case of initial zero residual in gmres

alexfikl:


inducer:
> initial zero residual in gmres

Lucky you! How come you even need GMRES? :rofl:

inducer:
Thanks!

alexfikl:
> Lucky you! How come you even need GMRES?

Special cased it in the meantime, but came across it when solving a two-phase Stokes problem. In the case where there's no viscosity jump, the density is just the jump in traction (which I used as an initial guess for GMRES), so it doesn't actually need a solve.

Thought it's worth a fix because before it would just spit out NANs and happily iterate to `maxiter`.

PR #46, opened 2021-01-06
Tensor Product Element Support

alexfikl:
Mostly just tried to see what happens, in particular with the quad refinement. 

Currently it does seem to [refine something](https://i.imgur.com/srfYhUf.jpg), but looks a bit weird. Haven't looked into it, but `_simplex_mapping_max_stretch_factor` is suspiciously named when used with squares.

- [x] Requires inducer/meshmode#261
- [x] Point `requirements.txt` back to `meshmode@main`

### DIFF
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -309,6 +309,9 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
 
     # }}}
 
+    def map_error_expression(self, expr):
+        raise RuntimeError(expr.message)
+
     def map_is_shape_class(self, expr):
         discr = self.places.get_discretization(
             expr.dofdesc.geometry, expr.dofdesc.discr_stage)
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -309,6 +309,23 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
 
     # }}}
 
+    def map_is_shape_class(self, expr):
+        discr = self.places.get_discretization(
+            expr.dofdesc.geometry, expr.dofdesc.discr_stage)
+
+        from pytools import is_single_valued
+        if not is_single_valued(type(grp.mesh_el_group) for grp in discr.groups):
+            # FIXME Conceivably, one could stick per-group bools into a DOFArray.
+            raise NotImplementedError(
+                    "non-homogeneous element groups are not supported")
+
+        from meshmode.mesh import _ModepyElementGroup
+        meg = discr.groups[0].mesh_el_group
+        if isinstance(meg, _ModepyElementGroup):
+            return isinstance(meg._modepy_shape, expr.shape)
+        else:
+            raise TypeError(f"element type not supported: '{type(meg).__name__}'")
+
     def exec_assign(self, actx: PyOpenCLArrayContext, insn, bound_expr, evaluate):
         return [(name, evaluate(expr))
                 for name, expr in zip(insn.names, insn.exprs)]
@@ -828,7 +845,6 @@ class GeometryCollection:
         try:
             discr = self._get_discr_from_cache(geometry, discr_stage)
         except KeyError:
-            from pytential import sym
             from pytential.qbx.refinement import _refine_for_global_qbx
 
             # NOTE: this adds the required discretizations to the cache
--- a/pytential/symbolic/primitives.py
+++ b/pytential/symbolic/primitives.py
@@ -94,6 +94,11 @@ DOF Description
 .. autoclass:: DOFDescriptor
 .. autofunction:: as_dofdesc
 
+Diagnostics
+^^^^^^^^^^^
+
+.. autoclass:: ErrorExpression
+
 .. _placeholders:
 
 Placeholders
@@ -438,6 +443,23 @@ class Expression(ExpressionBase):
     def make_stringifier(self, originating_stringifier=None):
         from pytential.symbolic.mappers import StringifyMapper
         return StringifyMapper()
+
+
+class ErrorExpression(Expression):
+    """An expression that, if evaluated, causes an error with the supplied
+    *message*.
+
+    .. automethod:: __init__
+    """
+    init_arg_names = ("message",)
+
+    def __init__(self, message):
+        self.message = message
+
+    def __getinitargs__(self):
+        return (self.message,)
+
+    mapper_method = intern("map_error_expression")
 
 
 def make_sym_mv(name, num_components):
@@ -1013,7 +1035,8 @@ def _mapping_max_stretch_factor(ambient_dim, dim=None, dofdesc=None,
     stretch_factor = If(IsShapeClass(mp.Simplex, dofdesc),
                         simplex_stretch_factor,
                         If(IsShapeClass(mp.Hypercube, dofdesc),
-                           hypercube_stretch_factor, 0)
+                           hypercube_stretch_factor,
+                           ErrorExpression("unknown reference element shape"))
                         )
 
     return cse(stretch_factor, "mapping_stretch_factor", cse_scope.DISCRETIZATION)
--- a/pytential/symbolic/primitives.py
+++ b/pytential/symbolic/primitives.py
@@ -136,6 +136,7 @@ Functions
 Discretization properties
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
+.. autoclass:: IsShapeClass
 .. autoclass:: QWeight
 .. autofunction:: nodes
 .. autofunction:: parametrization_derivative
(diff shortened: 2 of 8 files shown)


inducer:
> Haven't looked into it, but `_simplex_mapping_max_stretch_factor` is suspiciously named when used with squares.

Took a quick look, this shouldn't be simplex-specific at all. One good test is to stick a (relatively tightly) shape-regular mesh on a sphere and look at adjacent elements. If `_simplex_mapping_max_stretch_factor` isn't continuous across element interfaces, something is off.

inducer:
Update: `_simplex_mapping_max_stretch_factor` is *not* fine (after discussion with @alexfikl, just pasting the chat log because lazy)

> Kidding aside, it tries to divine some sort of "stretch factor" out of the reference-to-global mapping
> Turns out one desirable property that one might want is that it doesn't change if you "rotate through" the vertices of the triangle
> If you just use the bi-unit reference element for that (i.e. bi-unit ref tri -> global), the "stretch factor" thing you get fails that test
> alexfikl: Ah, that's why it was mapped to an equilateral triangle
> Andreas: Use equilateral, and everything is peachy
> right
> That should probably be better commented/documented than it is
> Happy fact is that it should be much easier for quads to pass that test
> Because 🥁 nothing changes when you rotate them through
> My guess is that biunit is probably just fine
> But equilateral is definitely dead wrong
> And scaling matters... so that if you have a tri and a quad of equal sizes next to each other, this scale factor thing says the same thing
> alexfikl: Yeah, but this needs some changes somewhere to do different things for different element classes

alexfikl:
@inducer This should be ready for a look now. It mostly adds two things to get quad tensor product support:
* a `_hypercube_mapping_max_stretch_factor` that pretty much follows the simplex case.
* a `_ShapeDiscretizationProperty` expression that takes a dict of `shape_name -> expr` to hack in different expressions for the `_quad_resolution` based on the shape. 

There are some `vtu` files in [test_source_refinement_3d.zip](https://github.com/inducer/pytential/files/7110410/test_source_refinement_3d.zip) where
* `-original` is the original mesh
* `-stage1` is after the stage 1 refinement
* `-stage2` is after the stage 2 refinement

It seems to be doing what I would expect it to do, i.e. lower the max value of `_quad_resolution` as it goes through the stages. At the end of the day it looks something like this

Screenshot_20210904_144518

pytential/symbolic/primitives.py
```diff
@@ -966,7 +988,29 @@ def _simplex_mapping_max_stretch_factor(ambient_dim, dim=None, dofdesc=None,
...
+def _hypercube_mapping_max_stretch_factor(ambient_dim, dim=None, dofdesc=None,
+        with_elementwise_max=True):
+    if dim is None:
+        dim = ambient_dim - 1
+
+    # NOTE: unlike in the simplex case, here we do not need to transform the
+    # reference element, as it already is nicely rotation invariant
+    pder_mat = first_fundamental_form(ambient_dim, dim=dim, dofdesc=dofdesc)
+    stretch_factors = [
+            cse(4.0 * sqrt(s), f"hypercube_mapping_singval_{i}")
```
alexfikl, line 1003: Besides the obvious hackery of `_ShapeDiscretizationProperty`, I'm not quite sure where this `4` comes from. Without adding a factor here, it wouldn't refine anything on the spheroid case. I imagine it has something to do with simplices vs quads in the actual magnitude (2 simplices make up a quad?), but no clue really.

Any ideas?

inducer: The 4 in the simplex case corrects for the fact that the reference element is [-1,1], which shows up twice in J^T J (in case that wasn't clear from the comment). It would also show up twice in the first fundamental form.

What this function should do is faithfully approximate the max "stretch factor" vs a unit-size reference element. Note that the simplex case has the four *inside* the square root, the one here is outside. I think a 2 outside or a 4 inside is justifiable.

inducer: (And if, based on that, it decides no stage-2 refinement is needed, then great.)

alexfikl: Went ahead and changed that to a factor of 4 inside, exactly like for the simplex case.

alexfikl:
Added a quad sphere test to
* `test_global_qbx::test_source_refinement_3d`
* `test_layer_pot_identity::test_identity_convergence` (for Green + Laplace only)

Let me know if you want any additional tests.

alexfikl:
I've tested `test_stokeslet_identity` and `test_stresslet_identity` with the quad code form this PR and it seems to be doing a lot better. This is exactly the same setup as in [current main](https://github.com/inducer/pytential/blob/f37481d39dfb230306e29e90a868d22c681fb546/test/test_stokes.py#L311-L514) but `source_ovsmp` was set to 4 instead of 8. First, for the Stokeslet this gives (just showing the `x` component)

h              | Triangle       | EOC   | h              | Square         | EOC
:--------------|:---------------|:------|:---------------|:---------------|:------
1.30152308e+00 | 1.44793598e-04 |       | 1.89083426e+00 | 1.11018623e-04 |            
6.59334266e-01 | 6.86173087e-05 | 1.10  | 9.98418078e-01 | 1.16014264e-06 | 7.14       
3.30517323e-01 | 3.06025393e-05 | 1.17  | 4.99837310e-01 | 1.17708784e-07 | 3.31 

Then, for the Stresslet, this gives (also just the `x` component)

h              | Triangle       | EOC    | h              | Square         | EOC
:--------------|:---------------|:-------|:---------------|:---------------|:------
1.30152308e+00 | 1.58839109e-02 |        | 1.89083426e+00 | 5.52832593e-03 |                
6.59334266e-01 | 1.62710279e-02 | -0.04  | 9.98418078e-01 | 1.02202398e-04 | 6.25           
3.30517323e-01 | 1.58837030e-02 | 0.03   | 4.99837310e-01 | 1.98585048e-05 | 2.37

All of these just ran a variation of
```
python test_stokes.py 'test_stokeslet_identity(_acf, partial(eid.QuadSpheroidTestCase, resolutions=[0, 1, 2]), visualize=True)'
``` 

This seems to match what I was getting in my tensor product code as well. Not quite sure what's killing for the simplex case, but here are the results for all the iterations [stokesidentity.zip](https://github.com/inducer/pytential/files/7272168/stokesidentity.zip). It contains the errors, solutions, curvature, etc. to stare at.

pytential/symbolic/primitives.py
```diff
@@ -539,6 +548,22 @@ def __getinitargs__(self):
         return (self.dofdesc,)
 
 
+class _ShapeDiscretizationProperty(DiscretizationProperty):
```
inducer, line 551: Not sure I'm loving this. We already have a conditional. Why not use that? All we'd need in addition is a "Is the reference element in this group a ..." predicate.

inducer (commented):
Thanks! Generally good, though I hope we can replace the improvised conditional with an actual one. :)

pytential/symbolic/primitives.py
```diff
@@ -232,46 +232,53 @@ class _NoArgSentinel:
...
     pass
 
 
-class DEFAULT_TARGET:  # noqa: N801
+class DEFAULT_TARGET:                   # noqa: N801
     pass
 
 
-class QBX_SOURCE_STAGE1:   # noqa: N801
+class QBX_SOURCE_PRE_STAGE:             # noqa: N801
```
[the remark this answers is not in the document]
alexfikl: Renamed in 856d288. Added an underscore for good measure.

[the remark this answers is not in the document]
alexfikl: I added a comment and renamed it in becdf0e.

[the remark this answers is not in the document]
alexfikl: Went ahead and remove this. Was only used in one test and easily worked around.

pytential/symbolic/primitives.py
alexfikl: That's a lot nicer! Added something like that in 51ec319.

pytential/symbolic/primitives.py
```diff
@@ -944,21 +959,63 @@ def _simplex_mapping_max_stretch_factor(ambient_dim, dim=None, dofdesc=None,
...
+    hypercube_stretch_factor = _hypercube_mapping_max_stretch_factor(
+        ambient_dim, dim, dofdesc=dofdesc,
+        with_elementwise_max=with_elementwise_max)
+
+    import modepy as mp
+    from pymbolic.primitives import If
+    stretch_factor = If(IsShapeClass(mp.Simplex, dofdesc),
+                        simplex_stretch_factor,
+                        If(IsShapeClass(mp.Hypercube, dofdesc),
+                           hypercube_stretch_factor, 0)
```
alexfikl, line 1015: Is there a nice way to go `else raise` in here? I tried adding a `var("__unreacheable")`, but the compiler was complaining about it :(

alexfikl: `Quotient(1, 0)`? :grin:

inducer: Maybe make an `Error(string)` symbol?

pytential/symbolic/execution.py
```diff
@@ -309,6 +309,24 @@ def map_common_subexpression(self, expr):
...
+    def map_error_expression(self, expr):
+        raise RuntimeError(expr.message)
+
+    def map_is_shape_class(self, expr):
+        discr = self.places.get_discretization(
+            expr.dofdesc.geometry, expr.dofdesc.discr_stage)
+
+        from pytools import is_single_valued
+        if not is_single_valued(type(grp.mesh_el_group) for grp in discr.groups):
+            raise TypeError("non-homogeneous element groups are not supported")
```
inducer, line 321:
```suggestion
            # FIXME
            raise NotImplementedError("non-homogeneous element groups are not supported")
```
Conceivably, one could stick `True/False`s into a `DOFArray`.

inducer (commented):
Thanks, LGTM! I'll apply the one suggestion.

pytential/symbolic/primitives.py
```diff
@@ -439,6 +439,18 @@ def make_stringifier(self, originating_stringifier=None):
         return StringifyMapper()
 
 
+class ErrorExpression(Expression):
```
inducer, line 442: Missing docs, missing from the docs.

pytential/symbolic/primitives.py
```diff
@@ -539,6 +552,26 @@ def __getinitargs__(self):
         return (self.dofdesc,)
 
 
+class IsShapeClass(DiscretizationProperty):
```
inducer, line 555: Missing from the docs.

inducer (commented):
Found two more things, will fix them.

PR #47, opened 2021-01-12
GitHub ci examples

inducer:

PR #48, opened 2021-01-12
Test #47 with numpy pinned

inducer:


inducer:
Nope. That wasn't it.

PR #55, opened 2021-02-03
Fix a bunch of deprecation warnings

alexfikl:
(just got numpy 1.20 locally)

https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

PR #56, opened 2021-02-04
Investigate CI failures

inducer:
cc @alexfikl

### DIFF
--- a/.test-conda-env-py3.yml
+++ b/.test-conda-env-py3.yml
@@ -5,7 +5,7 @@ channels:
 dependencies:
 - git
 - scipy
-- conda-forge::numpy < 1.20
+- conda-forge::numpy
 - conda-forge::sympy
 - pocl
 - islpy
--- a/.test-conda-env-py3.yml
+++ b/.test-conda-env-py3.yml
@@ -5,7 +5,7 @@ channels:
 dependencies:
 - git
 - scipy
-- conda-forge::numpy
+- conda-forge::numpy < 1.20
 - conda-forge::sympy
 - pocl
 - islpy
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ sympy
 git+https://github.com/inducer/modepy.git#egg=modepy
 git+https://github.com/inducer/pyopencl.git#egg=pyopencl
 git+https://github.com/inducer/islpy.git#egg=islpy
-git+https://github.com/inducer/loopy.git#egg=loopy
+git+https://github.com/inducer/loopy.git@fa1bda8ead922c93773c226e29c975df20c1be23#egg=loopy
 git+https://github.com/inducer/boxtree.git#egg=boxtree
 git+https://github.com/inducer/meshmode.git#egg=meshmode
 git+https://github.com/inducer/sumpy.git@3d99addef59a016e4a5802bab7e648cb8bf3006e#egg=sumpy
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,5 +8,5 @@ git+https://github.com/inducer/islpy.git#egg=islpy
 git+https://github.com/inducer/loopy.git#egg=loopy
 git+https://github.com/inducer/boxtree.git#egg=boxtree
 git+https://github.com/inducer/meshmode.git#egg=meshmode
-git+https://github.com/inducer/sumpy.git#egg=sumpy
+git+https://github.com/inducer/sumpy.git@3d99addef59a016e4a5802bab7e648cb8bf3006e#egg=sumpy
 git+https://github.com/inducer/pyfmmlib.git#egg=pyfmmlib


inducer:
Huh. Works like a charm. I'll rerun to see if it's a fluke.

alexfikl:
Seems to have failed again. Hm, so it's not numpy and not related to anything in #55.

inducer:
Thanks @isuruf for checking for sumpy influence!

I just added a run without

- https://github.com/inducer/loopy/pull/197
- https://github.com/inducer/loopy/pull/217

cc @kaushikcfd

kaushikcfd:
Hmm. `CI/examples` passed when an older loopy is pinned. I don't have a fix yet, but:
1. locally `laplce-dirichlet-3d.py` passes for me. (for both loopy versions)
2. iname tags of kernels generated by the 2 commits with discrepancy are identical.

inducer:
So the first go passed. Now running a second round to see if it's repeatable.

inducer:
For perspective, the examples failure is distinct. Full story at #57. This did not exhibit the failure I am concerned about here, which is the main Linux pytest failure. I'll rerun again, to see if it holds up.

alexfikl:
Did the Linux tests fail after numpy was bumped down? I just remember the examples failing in a while now.

Those failures really looked like some sort of out of memory issue. Did https://gitlab.tiker.net/inducer/pytential/-/issues/131 ever improve?

inducer:
The only way I know to curb this misery going forward is running downstream CI along with upstream projects, in this case loopy, as I propose here: https://github.com/inducer/loopy/pull/220. If that works out, I'll probably apply the same idea to meshmode.

inducer:
> Did the Linux tests fail after numpy was bumped down?

I don't recall such an instance. I'll bump numpy back up here, to check. But I don't expect it to fail.

> Those failures really looked like some sort of out of memory issue.

I agree, though I don't see (yet?) how the loopy PRs would inflate memory usage in a substantial fashion.

> Did https://gitlab.tiker.net/inducer/pytential/-/issues/131 ever improve?

No, didn't. It just wasn't bad enough to be a problem. In addition, there's a similar-looking mystery (https://github.com/illinois-ceesd/mirgecom/issues/212) being chased down in mirgecom.

alexfikl:
> I don't recall such an instance. I'll bump numpy back up here, to check. But I don't expect it to fail.

It seems to have failed, rerun?

> No, didn't. It just wasn't bad enough to be a problem. In addition, there's a similar-looking mystery (illinois-ceesd/mirgecom#212) being chased down in mirgecom.

Maybe worth adding a memory pool already in [pytest_generate_tests_for_pyopencl_array_context](https://github.com/inducer/meshmode/blob/91f070e87156854582140712cc7709cd5a505534/meshmode/array_context.py#L549)? Although yeah, that would just hide the issue.

inducer:
> It seems to have failed, rerun?

Wha? :exploding_head: 

Sure, I'll rerun, but now I don't know *what* to believe. Is this something that's brought about by numpy and the loopy change together?

inducer:
> Maybe worth adding a memory pool already in [pytest_generate_tests_for_pyopencl_array_context](https://github.com/inducer/meshmode/blob/91f070e87156854582140712cc7709cd5a505534/meshmode/array_context.py#L549)? Although yeah, that would just hide the issue.

Ugh, no. Not a fan of sweeping stuff under the rug.

alexfikl:
> Sure, I'll rerun, but now I don't know _what_ to believe. Is this something that's brought about by numpy and the loopy change together?

Just to add another variable: looking at the CI history, [last scheduled run](https://github.com/inducer/pytential/runs/1756229624) on Ubuntu 18.04 passed just fine, but then the [next](https://github.com/inducer/pytential/actions/runs/524551719) [ones](https://github.com/inducer/pytential/actions/runs/544474822) on Ubuntu 20.04 started failing. Can we pin it to Ubuntu 18.04 to see if that passes reliably?

Besides that, no idea, since it seems to pass intermittently..

inducer:
Hmm, so possibly the common theme among all these changes (newer ubuntu, loopy PRs, numpy 1.20) is just that they each ever so slightly increase memory usage...

inducer:
Passed this time around, FWIW.

inducer:
Alright, I'm now super confused. Reverting the Loopy PRs that we suspected caused problems actually did exactly *nothing* to help https://github.com/inducer/loopy/pull/220 pass. So that theory is pretty dead in the water to me.

inducer:
My next best plan is to go hunt this stupid memory leak. Grrr.

inducer:
https://github.com/illinois-ceesd/mirgecom/issues/212 if you'd like to follow the saga.

inducer:
Using jemalloc for the CI (#58) seems to help. See https://github.com/illinois-ceesd/mirgecom/issues/212 for more details. Closing here.

PR #38, opened 2020-11-15
IntG with multiple kernels

isuruf:
- [x] Requires inducer/sumpy#23
- [x] Point `requirements.txt` back to master

PR #66, opened 2021-03-04
Add missing source/target kernels to P2PMatrixBuilder

alexfikl:
I missed this while looking over #38, but it seems like it should be there, right?


inducer:
Yep, agreed. LGTM. Thanks for working on this!

isuruf:
Thanks for fixing it.

PR #29, opened 2020-09-10
Stokes biharmonic

isuruf:
Mirror of https://gitlab.tiker.net/inducer/pytential/-/merge_requests/163 for review

PR #71, opened 2021-04-06
Faster direct evaluation

alexfikl:
This is supposed to help a bit with Stokes until #29 and friends is in. It adds a few small things that seemed to help make it faster

* ~~A `flatten` symbolic function, which is mainly used as a way to cache the required flattened geometry inside the `GeometryCollection`.~~ Cached the flattened nodes, expansion radii and centers in the `GeometryCollection` using `memoize_in` directly.
* ~~Calling `BoundExpression` directly instead of `bind` because `_prepare_expr` showed up significantly. This is fine because in `exec_compute_potential_insn_direct` we already tag everything and don't have any `IntG`, so there's no need for the preprocessing.~~ (turns out this was a very small save, if any)
* In `BoundExpression`, make the call to `OperatorCompiler` lazy and add an early exist in `eval` when the expression is a `cse` scoped on the discretization. Not sure this helped much, but it made the tree in vmprof smaller!

The biggest save came from redoing the `for o in insn.outputs` loop. From what I understand, `sumpy.qbx.LayerPotential` can compute all the target kernels at once now, but that loop computed them all for each element in `insn.outputs` and then just took the one in `o.target_kernel_index` (effectively redoing the computations loads of times).

This probably still needs a bit of testing to make sure nothing is broken, but it seems to help quite a lot, i.e. it went from around 100s to around 10s on a benchmark script I was playing with.

TODO
- [x] Needs inducer/sumpy#41
- [x] Point `requirements.txt` back to sumpy master

### DIFF
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -303,6 +303,11 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
         arg, = args
         return abs(self.rec(arg))
 
+    def apply_flatten(self, args):
+        from pytential.utils import flatten_if_needed
+        arg, = args
+        return flatten_if_needed(self.array_context, self.rec(arg))
+
     # }}}
 
     def map_call(self, expr):
@@ -310,7 +315,7 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
                 EvalMapperFunction, NumpyMathFunction)
 
         if isinstance(expr.function, EvalMapperFunction):
-            return getattr(self, "apply_"+expr.function.name)(expr.parameters)
+            return getattr(self, f"apply_{expr.function.name}")(expr.parameters)
         elif isinstance(expr.function, NumpyMathFunction):
             args = [self.rec(arg) for arg in expr.parameters]
             from numbers import Number
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -210,7 +210,7 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
         if isinstance(x, DOFArray):
             return thaw(self.array_context, x)
         else:
-            return self.array_context.thaw(x)
+            return self.array_context.thaw(x).copy()
 
     def map_num_reference_derivative(self, expr):
         discr = self.places.get_discretization(
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -226,7 +226,8 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
         return thaw(self.array_context, discr.quad_weights())
 
     def map_inverse(self, expr):
-        bound_op_cache = self.bound_expr.places._get_cache("bound_op")
+        bound_op_cache = self.bound_expr.places._get_cache(
+                (EvaluationMapperBase, "bound_op"))
 
         try:
             bound_op = bound_op_cache[expr]
@@ -259,9 +260,9 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
 
     def map_common_subexpression(self, expr):
         if expr.scope == sym.cse_scope.EXPRESSION:
-            cache = self.bound_expr._get_cache("cse")
+            cache = self.bound_expr._get_cache((EvaluationMapperBase, "cse"))
         elif expr.scope == sym.cse_scope.DISCRETIZATION:
-            cache = self.places._get_cache("cse")
+            cache = self.places._get_cache((EvaluationMapperBase, "cse"))
         else:
             return self.rec(expr.child)
 
@@ -615,10 +616,6 @@ def _is_valid_identifier(name):
     return name.isidentifier() and not keyword.iskeyword(name)
 
 
-_GEOMETRY_COLLECTION_DISCR_CACHE_NAME = "refined_qbx_discrs"
-_GEOMETRY_COLLECTION_CONNS_CACHE_NAME = "refined_qbx_conns"
-
-
 class GeometryCollection:
     """A mapping from symbolic identifiers ("place IDs", typically strings)
     to 'geometries', where a geometry can be a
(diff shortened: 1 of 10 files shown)


alexfikl:
Should I add the testing script I have as an example? It mostly just sums up a lot of Stokes kernels and their (target) derivatives.

pytential/symbolic/primitives.py
```diff
@@ -509,6 +509,7 @@ def __call__(self, *args, **kwargs):
 imag = EvalMapperFunction("imag")
 conj = EvalMapperFunction("conj")
 abs = EvalMapperFunction("abs")
+flatten = EvalMapperFunction("flatten")
```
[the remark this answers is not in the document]
alexfikl: Done in bc633da.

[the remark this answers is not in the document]
alexfikl: Added one in 2efc517. 

It's not actually checked anywhere during the evaluation that the result is or isn't flattened, so this has a good chance of blowing up if a flattened expression is added in the middle of something.

pytential/symbolic/execution.py
```diff
@@ -210,7 +210,7 @@ def map_node_coordinate_component(self, expr):
         if isinstance(x, DOFArray):
             return thaw(self.array_context, x)
         else:
-            return self.array_context.thaw(x)
+            return self.array_context.thaw(x).copy()
```
inducer, line 213: Why is that copy here?

pytential/symbolic/execution.py
```diff
@@ -257,6 +257,10 @@ def map_interpolation(self, expr):
         else:
             raise TypeError("cannot interpolate `{}`".format(type(operand)))
 
+    def map_flatten(self, expr):
+        from pytential.utils import flatten_if_needed
+        return flatten_if_needed(self.array_context, self.rec(expr.operand))
```
inducer, line 262: This is too loose for my taste. I'd like it to fail if it's given something that's already flattened.

pytential/symbolic/primitives.py
```diff
@@ -1123,6 +1152,34 @@ def weights_and_area_elements(ambient_dim, dim=None, dofdesc=None):
...
+                dofdesc),
+            scope=cse_scope.DISCRETIZATION)
+
+
+def _flat_expansion_centers(ambient_dim, side, dim=None, dofdesc=None):
+    return cse(
+            _Flatten(
+                expansion_centers(ambient_dim, side, dim=dim, dofdesc=dofdesc),
+                dofdesc),
+            scope=cse_scope.DISCRETIZATION)
```
inducer, line 1178: I get *why* these exist, but they're annoying IMO. They inflict management of a concern on the user (flattening) that the infrastructure should be able to figure out without paying a massive penalty. If we made use of the fact that (in many common cases) flattening is a reshape, would that allow us to not do this and still dodge most of the flattening cost?

pytential/symbolic/execution.py
alexfikl: This case happens for `PointsTarget` and `PointPotentialSource` type things, which have stacked nodes of some shape ``(dim, nnodes)``. 

From what I can tell, the issue is a missing `default_offset=lp.auto` in [sumpy.p2p](https://github.com/inducer/sumpy/blob/b0ec769453fb8052da1cbd5a8ab6070e6015de52/sumpy/p2p.py#L216-L242), but I tried adding that and it had no effect. Any idea why that would be?

pytential/symbolic/primitives.py
alexfikl: I'm really starting to dislike this too, at least the symbolic part. I'll go ahead and remove it for now.

alexfikl: Removed in 1a2338d.

It seems to add about 1s (out of a total of 9s-ish) to the case I was playing with. I imagine that'll get worse for larger geometries / more groups. 

How do you feel about just adding a `memoize_in(self)` in `exec_compute_potential_insn_direct` for this? The FMM side seems to do its own flattening in the `GeometryData` classes, so it doesn't feel like it's useful to do something too general here.

inducer: I'm OK with that.

pytential/symbolic/execution.py
inducer: > I tried adding that and it had no effect. Any idea why that would be?

Did you clear/ignore caches after you added it? `SUMPY_NO_CACHE=1`.

inducer:
Unsubscribing... @-mention or request review once it's ready for a look or needs attention.

pytential/symbolic/execution.py
alexfikl: Yeah.. :facepalm: 

inducer/sumpy#41

pytential/symbolic/primitives.py
alexfikl: Added in 43a9d9a.

They're memoized in the geometry collection that's passed in.

pytential/symbolic/execution.py
```diff
@@ -1030,6 +1033,15 @@ def eval(self, context=None, timing_data=None,
         if context is None:
             context = {}
 
+        # NOTE: avoid compiling any code if the expression is long lived
+        # and already nicely cached in the collection from a previous run
+        import pymbolic.primitives as prim
+        if isinstance(self.sym_op_expr, prim.CommonSubexpression) \
+                and self.sym_op_expr.scope == sym.cse_scope.DISCRETIZATION:
+            cache = self.places._get_cache("cse")
```
inducer, line 1041: Use a specially-made class symbol here instead of "cse", to make clashes less likely.

inducer (commented):
Thanks! LGTM, just a few small things.

pytential/symbolic/execution.py
alexfikl: That's what's used in [EvaluationMapper.map_common_subexpression](https://github.com/inducer/pytential/blob/d3b25669c95335b515b7b5b97833f75a70f19b5b/pytential/symbolic/execution.py#L260-L274). Should I change it there too?

inducer: Thanks for noticing this. Yeah, I think so. I think that code was written before the message "symbols are unique-er than strings but just as hashable" sunk in with me.

alexfikl: Went ahead and updated this and other cache names I could find in 38a6290.

inducer: Pushed a counterproposal in 365a591. Let me know if you're OK with that.

alexfikl: Yep, looks good to me!

alexfikl:
Thanks for the review and merge! This is a pretty big quality-of-life improvement for my big Stokes expressions!

PR #73, opened 2021-04-22
Check matrix conditioning

alexfikl:
The condition number for `0.5 * u + D(u, "avg")` seems to be huge when constructing the matrix with `pytential.symbolic.execute.build_matrix`, but it's not for `-0.5 * u + D(u, "avg")`.

@inducer

### DIFF
--- a/test/test_matrix.py
+++ b/test/test_matrix.py
@@ -182,6 +182,11 @@ def test_build_matrix(ctx_factory, k, curve_fn, op_type, visualize=False):
 @pytest.mark.parametrize("side", [+1, -1])
 @pytest.mark.parametrize("op_type", ["single", "double"])
 def test_build_matrix_conditioning(ctx_factory, side, op_type, visualize=False):
+    """Checks that :math:`I + K`, where :math:`K` is compact gives a
+    well-conditioned operator when it should. For example, the exterior Laplace
+    problem has a nullspace, so we check that and remove it.
+    """
+
     ctx = ctx_factory()
     queue = cl.CommandQueue(ctx)
     actx = PyOpenCLArrayContext(queue)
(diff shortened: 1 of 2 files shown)


alexfikl:
Doing and SVD on the two gave this: `side=-1` (left) and `side=+1` (right)

<p align="center">


</p>

It does seem like it has a singular value dangerously close to 0 there. This seems to stay there when going from a circle to an ellipse, so it doesn't seem like it's a fluke (?).

inducer:
Oh right. Exterior Dirichlet Laplace has a nullspace, cf. [here](https://github.com/inducer/pytential/blob/094b75ba3ece5314b6ce6f7930d946fe5efbb8bb/pytential/symbolic/pde/scalar.py#L124-L127). Sorry, I should've remembered that when we chatted. Once you nudge the constant out of the nullspace (like [here](https://github.com/inducer/pytential/blob/094b75ba3ece5314b6ce6f7930d946fe5efbb8bb/pytential/symbolic/pde/scalar.py#L157-L167)), the conditioning should become entirely benign. So I don't think there's anything to see here, this is entirely expected.

inducer:
(It might not hurt to keep that test around though.)

alexfikl:
> (It might not hurt to keep that test around though.)

Yeah, sure. What should it test though? That there's a singular value close to zero in that one case?

inducer:
Just slap a reasonable upper bound on it? Basically... not 1e10? :)

alexfikl:
> Just slap a reasonable upper bound on it? Basically... not 1e10? :)

@inducer Added an assert like that and everything seems to pass, so it's good to go from my side.

test/test_matrix.py
```diff
@@ -179,6 +179,96 @@ def test_build_matrix(ctx_factory, k, curve_fn, op_type, visualize=False):
...
+        ))
+
+    kappa = la.cond(mat)
+    _, sigma, _ = la.svd(mat)
+
+    logger.info("cond: %.5e sigma_max %.5e", kappa, sigma[0])
+
+    # NOTE: exterior Laplace has a nullspace
+    if side == +1 and op_type == "double":
+        assert kappa > 1.0e+9
```
inducer, line 241: Oh, sorry! What I meant was that we should test that the matrix with "ones(n, n)" added to it should have benign conditioning, as in the IE test. (This removes the nullspace.)

alexfikl: Added in 7e78c22. Better?

inducer: Yup, thanks!

PR #74, opened 2021-04-27
Docs and cleanup for scalar operators

alexfikl:
Wanted to use some of this and ended up adding some docs and making the interfaces a bit more similar in the process.

### DIFF
--- a/pytential/symbolic/pde/scalar.py
+++ b/pytential/symbolic/pde/scalar.py
@@ -42,6 +42,9 @@ from sumpy.kernel import DirectionalSourceDerivative
 
 class L2WeightedPDEOperator:
     """
+    .. attribute:: kernel
+    .. attribute:: use_l2_weighting
+
     .. automethod:: get_weight
     .. automethod:: get_sqrt_weight
 
@@ -68,6 +71,11 @@ class L2WeightedPDEOperator:
         return self.kernel.dim
 
     def get_weight(self, dofdesc=None):
+        """
+        :returns: a symbolic expression for the weights (quadrature weights
+            and area elements) on *dofdesc* if :attr:`use_l2_weighting` is
+            *True* and ``1`` otherwise.
+        """
         if self.use_l2_weighting:
             return sym.cse(
                     sym.area_element(self.dim, dofdesc=dofdesc)
@@ -76,25 +84,42 @@ class L2WeightedPDEOperator:
             return 1
 
     def get_sqrt_weight(self, dofdesc=None):
+        """
+        :returns: the square root of :meth:`get_weight`.
+        """
         if self.use_l2_weighting:
             return sym.sqrt_jac_q_weight(self.dim, dofdesc=dofdesc)
         else:
             return 1
 
     def prepare_rhs(self, b):
+        """Modify the right-hand side (e.g. boundary conditions) to match the
+        operator constructed in :meth:`operator`.
+        """
         return self.get_sqrt_weight() * b
 
     def get_density_var(self, name: str):
         """
-        :returns: a symbolic variable/array corresponding to the density
-            with the given *name*.
+        :param name: a string name for the density.
+        :returns: a symbolic variable or array (problem dependent)
+            corresponding to the density with the given *name*.
         """
         return sym.var(name)
 
-    def representation(self, u, qbx_forced_limit=None):
+    def representation(self, u, qbx_forced_limit=None, **kwargs):
+        """
+        :returns: a representation for the unknowns based on an integral
+            equation with density *u*. If *qbx_forced_limit* denotes an
+            on-surface evaluation, the corresponding jump relations are not
+            added to the representation.
+        """
         raise NotImplementedError
 
-    def operator(self, u):
+    def operator(self, u, **kwargs):
+        """
+        :returns: a boundary integral operator with corresponding jump
+            relations that can be used to solve for the density *u*.
+        """
         raise NotImplementedError
 
 # }}}
@@ -116,6 +141,9 @@ class DirichletOperator(L2WeightedPDEOperator):
     Inherits from :class:`L2WeightedPDEOperator`.
 
     .. automethod:: is_unique_only_up_to_constant
+    .. automethod:: representation
+    .. automethod:: operator
+    .. automethod:: __init__
     """
 
     def __init__(self, kernel: Kernel, loc_sign: int, *,
@@ -136,7 +164,7 @@ class DirichletOperator(L2WeightedPDEOperator):
 
         if alpha is None:
             from sumpy.kernel import LaplaceKernel
-            if isinstance(self.kernel, LaplaceKernel):
+            if isinstance(kernel, LaplaceKernel):
                 alpha = 0
             else:
                 alpha = 1j
@@ -156,7 +184,7 @@ class DirichletOperator(L2WeightedPDEOperator):
             map_potentials=None, qbx_forced_limit=None, **kwargs):
         """
         :param u: symbolic variable for the density.
-        :param map_potentials: a :class:`Callable` that can be used to apply
+        :param map_potentials: a callable that can be used to apply
             additional transformations on all the layer potentials in the
             representation, e.g. to take a target derivative.
         :param kwargs: additional keyword arguments passed on to the layer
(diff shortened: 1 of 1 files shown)


pytential/symbolic/pde/scalar.py
```diff
@@ -48,110 +66,157 @@ def __init__(self, kernel, use_l2_weighting):
...
-
-        L2WeightedPDEOperator.__init__(self, kernel, use_l2_weighting)
 
         if alpha is None:
-            if isinstance(self.kernel, LaplaceKernel):
+            from sumpy.kernel import LaplaceKernel
+            if isinstance(kernel, LaplaceKernel):
                 alpha = 0
             else:
                 alpha = 1j
```
alexfikl, line 170: I was trying to use this for Yukawa and wasn't sure what `alpha` should be. Is this `1j` meant for Helmholtz?

pytential/symbolic/pde/scalar.py
```diff
@@ -162,86 +227,104 @@ def operator(self, u):
...
 
     .. automethod:: is_unique_only_up_to_constant
     .. automethod:: representation
     .. automethod:: operator
+    .. automethod:: __init__
     """
 
-    def __init__(self, kernel, loc_sign, alpha=None,
-            use_improved_operator=True,
-            laplace_kernel=0, use_l2_weighting=False,
```
alexfikl, line 204: I removed this `laplace_kernel=0` since it wasn't used and wasn't cleanr what it was for. Any ideas?

pytential/symbolic/pde/scalar.py
```diff
@@ -252,108 +335,139 @@ def map_potentials(x):  # pylint:disable=function-redefined
...
-                    qbx_forced_limit="avg")
+            def Sp(density):
+                return sym.Sp(self.laplace_kernel,
+                        density,
+                        qbx_forced_limit="avg")
+
+            Dp0S0u = -0.25 * u + Sp(Sp(inv_sqrt_w_u))
 
+            from sumpy.kernel import HelmholtzKernel, LaplaceKernel
             if isinstance(self.kernel, HelmholtzKernel):
```
alexfikl, line 368: Do you happen to have any reference for this? Is it in that Hackbusch book?

inducer: It's just "right-precondition with a single layer and use a Calderón identity", see e.g. [here](https://andreask.cs.illinois.edu/cs598apk-f19/notes.pdf#page=278). The point is that the "improved" `Dp0S0u` is second-kind. But yes, I agree that this deserves a comment. :)

pytential/symbolic/pde/scalar.py
inducer: `alpha = 0` is the right choice for Yukawa. (That conditional should be fixed to be Helmholtz-only.) The point of `alpha` is the [Brakhage-Werner trick](https://andreask.cs.illinois.edu/cs598apk-f19/notes.pdf#page=270) to get rid of spurious resonances for Helmholtz.

inducer: Probably also deserves a comment. :)

pytential/symbolic/pde/scalar.py
inducer: :shrug: Backward compatibility with some braindeadness? I'm sure that can go.

inducer:
Thanks for working on this. Ping me when it's good to go.

pytential/symbolic/pde/scalar.py
alexfikl: Added some comments and flipped the if in d52cd68

pytential/symbolic/pde/scalar.py
alexfikl: Added some notes in 26d3c7f.

pytential/symbolic/pde/scalar.py
```diff
@@ -252,108 +342,147 @@ def map_potentials(x):  # pylint:disable=function-redefined
...
+            #   D' S = -u/4 + S'^2
+            Dp0S0u = -0.25 * u + Sp(Sp(inv_sqrt_w_u))
+
+            from sumpy.kernel import HelmholtzKernel, LaplaceKernel
             if isinstance(self.kernel, HelmholtzKernel):
-                DpS0u = (  # noqa
-                        sym.Dp(knl - lknl,  # noqa
-                            cse(sym.S(lknl, inv_sqrt_w_u, qbx_forced_limit=+1)),
-                            qbx_forced_limit=+1, **knl_kwargs)
+                DpS0u = (
```
alexfikl, line 384: Shouldn't the same trick work for Yukawa? i.e. adding a `Dp(self.kernel - self.laplace_kernel)` to subtract out the added Laplace stuff. 

Although if `alpha = 0` is recommended for Yukawa, this makes no difference.

pytential/symbolic/pde/scalar.py
```diff
@@ -27,19 +27,37 @@
...
+    .. automethod:: get_sqrt_weight
+
+    .. automethod:: get_density_var
+    .. automethod:: prepare_rhs
+
+    .. automethod:: representation
+    .. automethod:: operator
+
+    .. automethod:: __init__
+    """
```
inducer, line 58:
```suggestion
    :math:`L^2`-weighting is performed to help with the solution of IEs that yield
    locally singular densities. It does this by matching the :math:`\ell^2`-norm used
    by the iterative method (e.g. GMRES) with the (approximate) :math:`L^2`-norm.
    Clearly, singular densities might not have finite :math:`\ell^2`-norm, hampering
    convergence of the iterative method. See
    `Bremer, J. On the Nyström discretization of integral equations on planar
    curves with corners. ACHA, 2011. <https://doi.org/10.1016/j.acha.2011.03.002>`__
    """
```

pytential/symbolic/pde/scalar.py
```diff
@@ -48,110 +66,160 @@ def __init__(self, kernel, use_l2_weighting):
...
-
-        self.loc_sign = loc_sign
-
-        L2WeightedPDEOperator.__init__(self, kernel, use_l2_weighting)
 
         if alpha is None:
-            if isinstance(self.kernel, LaplaceKernel):
-                alpha = 0
-            else:
+            # See Culton and Kress Chapter 3 for an explanation
```
inducer, line 168:
```suggestion
            # See Coulton and Kress Chapter 3 for an explanation
```

inducer, line 168:
```suggestion
            # Use a combined-field/Brakhage-Werner representation
            # (alpha != 0) to avoid spurious resonances (mainly for
            # the exterior problem)
            # See:
            # - Brakhage and Werner.
            #    Über das Dirichletsche Außenraumproblem für die
            #    Helmholtzsche Schwingungsgleichung.
            #    https://doi.org/10.1007/BF01220037
            # - Coulton and Kress, Chapter 3
```

pytential/symbolic/pde/scalar.py
```diff
@@ -162,74 +230,89 @@ def operator(self, u):
...
-        self.loc_sign = loc_sign
-        self.laplace_kernel = LaplaceKernel(kernel.dim)
-
-        L2WeightedPDEOperator.__init__(self, kernel, use_l2_weighting)
 
         if alpha is None:
-            if isinstance(self.kernel, LaplaceKernel):
-                alpha = 0
-            else:
+            # See Culton and Kress Chapter 3 for an explanation
```
inducer, line 303:
```suggestion
            # See above (Ctrl-F Brakhage-Werner)
```

pytential/symbolic/pde/scalar.py
inducer: It would, but there's no point to the combined-field representation for Yukawa. It'll work, but there's really no need.

inducer (commented):
Thanks! A few FYIs suggestions for you to look over, otherwise LGTM.

pytential/symbolic/pde/scalar.py
alexfikl: The internet says it's Colton
https://www.springer.com/gp/book/9781461449416

inducer: Argh, you're right!

inducer:
Thanks!

PR #77, opened 2021-04-30
Failing matrix evaluation with target derivatives

alexfikl:
This now triggers the exception from [sumpy](https://github.com/inducer/sumpy/blob/fbd0b8e18dde5ee1350b38d844464565c863ce2f/sumpy/expansion/local.py#L107-L109). Not quite sure what the right fix is, but there's a `LineTaylorLocalExpansion` in the [matrix evaluation](https://github.com/inducer/pytential/blob/062dd89597b1b1ae2d4bcf8b4c6db1db72ac7466/pytential/symbolic/matrix.py#L381-L383).

@inducer Suggestions?

### DIFF
--- a/test/test_matrix.py
+++ b/test/test_matrix.py
@@ -51,6 +51,51 @@ logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
 
 
+def test_target_derivative_sumpy(ctx_factory):
+    cl_ctx = ctx_factory()
+    queue = cl.CommandQueue(cl_ctx)
+    actx = PyOpenCLArrayContext(queue)
+
+    # prevent cache explosion
+    from sympy.core.cache import clear_cache
+    clear_cache()
+
+    case = extra.CurveTestCase(
+            name="ellipse",
+            curve_fn=lambda t: ellipse(3.0, t),
+            target_order=4,
+            qbx_order=4,
+            resolutions=[64],
+            )
+    logger.info("\n%s", case)
+
+    # {{{ check
+
+    qbx = case.get_layer_potential(actx, case.resolutions[-1], case.target_order)
+
+    from pytential.qbx.refinement import refine_geometry_collection
+    places = GeometryCollection(qbx, auto_where=case.name)
+    places = refine_geometry_collection(places,
+            refine_discr_stage=sym.QBX_SOURCE_QUAD_STAGE2)
+
+    from sumpy.kernel import LaplaceKernel
+    knl = LaplaceKernel(places.ambient_dim)
+
+    sym_u = sym.var("u")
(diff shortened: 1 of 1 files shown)


inducer:
Did line Taylor with target derivatives ever give the right answer?

alexfikl:
> Did line Taylor with target derivatives ever give the right answer?

No idea really. The only tests here check that it gave the same result as the matrix free evaluation. There are no matrix tests that actually solve something with `Sp`.

inducer:
I think it's highly unlikely that it was ever computing the right thing, so in that sense, the error message is right. A not totally unreasonable workaround is to just use a volume expansion in that situation.

alexfikl:
> I think it's highly unlikely that it was ever computing the right thing, so in that sense, the error message is right. A not totally unreasonable workaround is to just use a volume expansion in that situation.

But then it wouldn't match what `QBXLayerPotential` is doing for the direct eval in [here](https://github.com/inducer/pytential/blob/062dd89597b1b1ae2d4bcf8b4c6db1db72ac7466/pytential/qbx/__init__.py#L701-L705)?

inducer:
Yep, on-surface direct eval using line Taylor is just wrong for anything with target derivatives. For example, using the Calderon identity test in `test_identity_convergence`, modified to use direct eval, and after disabling the exception in sumpy:
```
---> (5, '5-starfish-0.25', 30, 'calderon') 0.4800855751905867
---> (5, '5-starfish-0.25', 50, 'calderon') 0.4800084963924671
---> (5, '5-starfish-0.25', 70, 'calderon') 0.4801550821046587
---> (5, '5-starfish-0.25', 90, 'calderon') 0.4802059307056389
h                   | Error              | Running EOC           
--------------------+--------------------+------------------------
0.33786640504191023 | 0.4800855751905867 |                       
0.20267565679346258 | 0.4800084963924671 | 0.00031419059202125624
0.14483907298023307 | 0.4801550821046587 | -0.0009087789244642336
0.11265893728817018 | 0.4802059307056389 | -0.000421457798346528 

Overall EOC: -0.00025374206741395815
```

alexfikl:
> Yep, on-surface direct eval using line Taylor is just wrong for anything with target derivatives. For example, using the Calderon identity test in `test_identity_convergence`, modified to use direct eval, and after disabling the exception in sumpy:
>
> `Overall EOC: -0.00025374206741395815`

What's wrong with some zero order convergence? At least you know where you are with it.. :facepalm: 

So the right fix would be to just use a volume expansion in both cases?

inducer:
#78 should address this.

alexfikl:
I'll go ahead and close this, since #78 should fail if the matrix parts don't work too.

PR #78, opened 2021-05-01
Support, test target derivatives for direct evaluation

inducer:
cc @alexfikl

### DIFF
--- a/pytential/qbx/__init__.py
+++ b/pytential/qbx/__init__.py
@@ -699,8 +699,19 @@ class QBXLayerPotentialSource(LayerPotentialSourceBase):
 
         from sumpy.qbx import LayerPotential
         from sumpy.expansion.local import LineTaylorLocalExpansion
+        from sumpy.kernel import TargetDerivativeRemover
+
+        tdr = TargetDerivativeRemover()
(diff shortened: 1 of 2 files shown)


alexfikl:
Thanks for fixing this!

Should the same chunk be added to the matrix evaluation? All of that is direct too.

inducer:
Yeah! (Is there a way that they could share that code?)

alexfikl:
> Yeah! (Is there a way that they could share that code?)

Maybe add a memoized `get_expansion_for_target_kernels` (or something?) with this logic to `QBXLayerPotential` so that the matrix code can just call that?

inducer:
That sounds good. `get_expansion_for_qbx_direct_eval` maybe? Feel free to push that to this PR if you like.

inducer:
Closing in favor of #79.

inducer:
Btw you're right: CI for this did eventually fail. :)

alexfikl:
> Btw you're right: CI for this did eventually fail. :)

I'm surprised this didn't already fail in inducer/sumpy#47 though. Was it not using the updated code for some reason?

PR #79, opened 2021-05-01
Add target derivative support to direct eval, matrix builders

alexfikl:
~~Targets #78~~ and seems to pass locally.


inducer:
Thanks! To avoid waiting for CI twice, I've closed #78 and changed the base branch here to `main`.

PR #80, opened 2021-05-05
Don't use rscale in qbx direct eval

isuruf:


isuruf:
`LineTaylorLocalExpansion` also ignores `rscale`

pytential/qbx/__init__.py
```diff
@@ -695,7 +695,7 @@ def get_expansion_for_qbx_direct_eval(self, base_kernel, target_kernels):
         tdr = TargetDerivativeRemover()
         if any(knl != tdr(knl) for knl in target_kernels):
             return self.expansion_factory.get_local_expansion_class(
-                    base_kernel)(base_kernel, self.qbx_order)
+                    base_kernel)(base_kernel, self.qbx_order, use_rscale=False)
```
inducer, line 698: Is there a test that could help guard against a regression?

inducer: Do you have a sense why `rscale` was a problem in the first place? `sumpy.qbx` supplies [the same value](https://github.com/inducer/sumpy/blob/3a8e79c10ed2f56f0765c4fca2711764145d169c/sumpy/qbx.py#L125-L127) of rscale to expand and evaluate, so that should be OK IMO.

I'm OK with converting this to a FIXME if it's not easy to fix now.

inducer (commented):
Thanks for identifying the issue! Two questions below.

isuruf:
Nvm, this was not the issue.

PR #86, opened 2021-05-22
Update deprecated calls to arraycontext

alexfikl:


pytential/symbolic/execution.py
```diff
@@ -83,7 +83,7 @@ def mesh_el_view(mesh, group_nr, global_array):
 
 
 class EvaluationMapperBase(PymbolicEvaluationMapper):
-    def __init__(self, bound_expr, actx: PyOpenCLArrayContext, context=None,
+    def __init__(self, bound_expr, actx: ArrayContext, context=None,
```
inducer, line 86: That was intentional. It really only does support `PyOpenCLArrayContext` for now. Same below.

Hopefully a `sed -i s/actx: ArrayContext/actx: PyOpenCLArrayContext` can mostly take care of this.

inducer (commented):
Thanks for doing this! I found one small problem, otherwise LGTM!

pytential/symbolic/execution.py
alexfikl: Yeah, got through about half of them and realized they're probably on purpose. Sorry for the noise!

Reverted in cdb0253.

alexfikl:
@inducer There seems to be a CL version/library mismatch in the mac CI, which is probably not caused by this PR.

inducer:
I think that was temporary, while pyopencl 2021.2.2 was propagating to the repos in conda forge.

pytential/symbolic/execution.py
inducer: Np at all. Thanks for helping with this!

inducer:
Thx!

inducer:
The CDN is probably still updating. https://github.com/conda-forge/pyopencl-feedstock/pull/67 only went in 3 hours ago. So it's deciding to compile from source since the conda-installed version is obsolete. The pyopencl source build (understandably) doesn't work on mac: It sees the headers from conda but uses the CL implementation from the OS. https://github.com/inducer/ci-support/commit/48067a9681d6a71ee7c0f5494e199f03dda7616f should fix that.

PR #30, opened 2020-09-23
Direct Solver: Skeletonization

alexfikl:
Attempting to break up the direct solver MR into smaller, more reviewable pieces. As a rundown

* **[Done]** Add proxy generation (already in ``linalg.proxy``).
* **[Current]** Add single-level skeletonization.
* **[TODO]** Add multi-level skeletonization.
* **[TODO]** Add the actual multi-level direct solver.

Previously found https://gitlab.tiker.net/inducer/pytential/-/merge_requests/211. 
The code for this is mostly in the deprecated https://gitlab.tiker.net/inducer/pytential/-/merge_requests/137.

PR #90, opened 2021-06-03
Improve proxy generator

alexfikl:
This takes out another chunk out of #30 that just changes the proxy generator. The bigger changes are
* Separated out a `ProxyGenerator`, that ignores any QBX expansions, and a `QBXProxyGenerator`, that makes proxies outside of the QBX expansions.
* Separated the loopy kernels that get centers and radii to make the proxy balls. Also split some inames and things, but that probably requires a careful (and knowledgeable) look.
* removed unused `gather_block_interaction_points`, which was meant to just make a big array of neighbors + proxies to be evaluated together.
* cleaned up the tests a bit and added some more checks.

Looking at the diff, it looks a bit big, but it should just be moving things around.

### DIFF
--- a/pytential/linalg/proxy.py
+++ b/pytential/linalg/proxy.py
@@ -20,6 +20,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
+from dataclasses import dataclass
 from typing import Optional
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


inducer:
That diff is genuinely unhelpful, but the end result looks OK. As for your `split_inames`: It wouldn't really fly on a GPU, but it'll do for CPU use. :)

inducer:
Thanks!

alexfikl:
> As for your `split_inames`: It wouldn't really fly on a GPU, but it'll do for CPU use. :)

I mostly copied that from some stuff I managed to sneak by you a long time ago in [sumpy](https://github.com/inducer/sumpy/blob/29a2fbd5d83fed8598fbaede1063c9145906f322/sumpy/qbx.py#L441). What would be a better way to optimize those kernels a bit?

inducer:
Got a whole course on that: https://relate.cs.illinois.edu/course/cs598apk-f18/ But I'd be happy to chat on Thursday. Which ones are the expensive ones?

alexfikl:
> Got a whole course on that: [relate.cs.illinois.edu/course/cs598apk-f18](https://relate.cs.illinois.edu/course/cs598apk-f18/) :) But I'd be happy to chat on Thursday. Which ones are the expensive ones?

It was just meant to do a bare minimum sort of thing, since I haven't profiled anything. For the moment, all the geometries we've been playing with have like 2-3000 points, so it probably doesn't matter that much anyway.

Thanks! I'll give that a read!

PR #91, opened 2021-06-10
Fix some loopy deprecation warnings

alexfikl:
Loopy complained that some of the instructions had to have some inames added.

```
venv/loopy/kernel/tools.py:271: LoopyWarning: in kernel compute_block_centers_knl: The iname(s) 
'irange' on instruction 'insn_1' was/were automatically added. This is deprecated. Please add 
the iname to the instruction explicitly, e.g. by adding 'for' loops (add 'inferred_iname'
 to silenced_warnings kernel attribute to disable)

```


inducer:
Thanks!

PR #92, opened 2021-06-11
Import and update BlockIndexRanges from sumpy

alexfikl:
These are removed from sumpy in inducer/sumpy#65.

This PR also reworks `BlockIndexRanges` a bit. Before the indices and ranges were `cl.array.Array`, which resulted in a lot of back and forth (when skeletonizing in #30), so now they're just `ndarray`. The arrays passed on to sumpy are still constructed as `cl.array.Array`.

- [x] Depends on inducer/sumpy#65
- [x] Point `requirements.txt` back to main sumpy.

### DIFF
--- a/doc/linalg.rst
+++ b/doc/linalg.rst
@@ -4,6 +4,12 @@ Linear Algebra Routines
 Hierarchical Direct Solver
 --------------------------
 
+.. warning::
+
+    All the classes and routines in this module are experimental and the
+    API can change at any point.
+
 .. automodule:: pytential.linalg.proxy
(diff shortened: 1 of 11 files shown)


inducer (approved):
Thanks!

inducer:
I'm assuming this is ready to un-draft, because otherwise thing would be left in a broken state. If not, stop me! :)

alexfikl:
> I'm assuming this is ready to un-draft, because otherwise thing would be left in a broken state. If not, stop me! :)

Yeah, should be good to go! Just needed inducer/sumpy#65.

PR #94, opened 2021-06-16
Fix QBX proxy radius

alexfikl:


pytential/linalg/proxy.py
```diff
@@ -463,7 +460,7 @@ def prg():
                          {dup=idim}
                 <> rqbx = rqbx_int if rqbx_ext < rqbx_int else rqbx_ext
 
-                proxy_radius[irange] = radius_factor * rblk
+                proxy_radius[irange] = radius_factor * rqbx
```
inducer, line 463: I feel like I'm lacking some context here, but it looks like this was previously computing and ignoring `rqbx`. Curious to hear if that had any impact on the direct solver error mystery we've been chasing.

Also looks like Loopy could/should have warned about this.

inducer (approved):
LGTM. Nice job finding this!

pytential/linalg/proxy.py
inducer: https://github.com/inducer/loopy/issues/423

PR #98, opened 2021-06-28
Update tests for device scalars

alexfikl:
This updates the tests and examples to
* use `meshmode.array_context.PyOpenCLArrayContext` instead of the one from `arraycontext`
* use the new `pytest_generate_tests_for_array_contexts` with `force_device_scalars=True`.
* use `actx_factory` instead of `ctx_factory` to get better defaults.

It's currently a draft because 
* [ ] It requires some resolution on how to broadcast device scalars with `cl.array.Array`.
* [x] Still need to go through kernels and add some tags.

### DIFF
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -186,10 +186,14 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
         return result
 
     def _map_elementwise_reduction(self, reduction_name, expr):
+        import loopy as lp
+        from arraycontext import make_loopy_program
+        from meshmode.transform_metadata import (
+                ConcurrentElementInameTag, ConcurrentDOFInameTag)
+
         @memoize_in(self.places, "elementwise_node_"+reduction_name)
         def node_knl():
-            from arraycontext import make_loopy_program
-            return make_loopy_program(
+            t_unit = make_loopy_program(
                     """{[iel, idof, jdof]:
                         0<=iel<nelements and
                         0<=idof, jdof<ndofs}""",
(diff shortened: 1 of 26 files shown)


inducer:
Thanks for going through this! I'd be all for merging this soon, even with the device scalar aspect half-resolved, to avoid having this bitrot.

alexfikl:
> Thanks for going through this! I'd be all for merging this soon, even with the device scalar aspect half-resolved, to avoid having this bitrot.

That's fine by me! 

I added some nasty hacks in `execution.py` that cast to `Number` anything that looks like a scalar array to convince the broadcasting to go through. Those probably require a careful look, but the rest is mostly search and replace.

pytential/symbolic/execution.py
```diff
@@ -124,31 +162,38 @@ def map_min(self, expr):
...
+        return result
 
     def map_node_max(self, expr):
-        # FIXME: make less CL specific
-        queue = self.array_context.queue
-        return max(
-                cl.array.max(grp_ary, queue=queue).get()[()]
-                for grp_ary in self.rec(expr.operand))
+        actx = self.array_context
+        result = max(actx.np.max(grp_ary) for grp_ary in self.rec(expr.operand))
```
inducer, line 174: TIL: `min`, `max` work on device scalars. I imagine it's this:
https://github.com/inducer/pyopencl/blob/a3d0b5cf4aceafa849e23486ace62962d5a2543f/pyopencl/array.py#L1467-L1472
that's doing that.

inducer (approved):
Took a look, looks great. No complaints.

inducer:
Thanks for taking this on!

pytential/symbolic/execution.py
alexfikl: Hm, that forces a bunch of `.get()`. Should it use `reduce(actx.np.maximum, ...)` instead?

inducer: It does. It's not ideal, but I can live with it for now. It's easy to fix later on when needed.

PR #99, opened 2021-06-30
Remove unneeded EvalMapperFunction

alexfikl:
This doesn't seem to be needed anymore because the array context vectorizes over object arrays already.

- [x] needs inducer/arraycontext#54
- [x] revert changes to `requirements.txt`


alexfikl:
Note to self for later: the tests fail because `actx.np.abs` returns complex numbers for complex inputs (with a zero imaginary part though).

inducer:
Thx!

PR #102, opened 2021-07-04
Pass translation_classes_data to sumpy

isuruf:
When this is passed to sumpy, it will use the optimized code path where the M2L matrix is precomputed.

### DIFF
--- a/pytential/qbx/__init__.py
+++ b/pytential/qbx/__init__.py
@@ -603,6 +603,9 @@ class QBXLayerPotentialSource(LayerPotentialSourceBase):
                     geo_data.tree().user_source_ids,
                     insn.kernel_arguments, evaluate))
 
+        from sumpy.fmm import SumpyTranslationClassesData
+        translation_classes_data = SumpyTranslationClassesData(actx.queue,
+                geo_data.traversal())
         tree_indep = self._tree_indep_data_for_wrangler(
                 target_kernels=insn.target_kernels,
                 source_kernels=insn.source_kernels)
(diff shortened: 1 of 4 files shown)


inducer:
Marking draft and unsubscribing for CI fails, let me know (request review/@-mention) when ready.

isuruf:
@inducer, this is ready.

pytential/qbx/__init__.py
```diff
@@ -603,6 +603,9 @@ def exec_compute_potential_insn_fmm(self, actx: PyOpenCLArrayContext,
                     geo_data.tree().user_source_ids,
                     insn.kernel_arguments, evaluate))
 
+        from sumpy.fmm import SumpyTranslationClassesData
+        translation_classes_data = SumpyTranslationClassesData(actx.queue,
+                geo_data.traversal())
```
inducer, line 608: This isn't a sumpy-specific interface. How come we're passing ostensibly sumpy-specific data?

pytential/qbx/__init__.py
```diff
@@ -603,6 +603,9 @@ def exec_compute_potential_insn_fmm(self, actx: PyOpenCLArrayContext,
...
         tree_indep = self._tree_indep_data_for_wrangler(
```
inducer, line 609: Would it make sense to compute the translation classes as part of the wrangler creation?

inducer (commented):
Thanks for working on this. Two questions below.

pytential/qbx/__init__.py
isuruf: Wrangler doesn't have access to a `queue`. Should I add that?

inducer: You can just make a private, temporary queue:
```python
with cl.CommandQueue(ctx) as queue:
   ...
```

isuruf:
Made redundant by https://github.com/inducer/sumpy/pull/117

PR #103, opened 2021-07-07
Disable N818

alexfikl:
There are two exceptions that make it complain: `NoInstructionAvailable` and `QBXTargetAssociationFailedException`. Would you prefer they were renamed?


inducer:
Not a fan of N818 (what were they thinking?!), so this LGTM. Thanks!

PR #104, opened 2021-07-07
Remove hacks for device scalars

alexfikl:
This should "just work" now that inducer/pyopencl#502 and inducer/arraycontext#51 are in.

### DIFF
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -100,44 +100,6 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
 
         self.queue = actx.queue
 
-    # {{{ TODO: remove when device scalar broadcasting is fixed
-
-    # NOTE:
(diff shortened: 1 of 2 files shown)


pytential/symbolic/execution.py
```diff
@@ -544,23 +500,19 @@ def matvec(self, x):
...
             assert x.shape == (self.total_dofs,)
-        elif isinstance(x, np.ndarray) and x.dtype.char == "O":
-            flat = False
-            host = False
-        elif isinstance(x, DOFArray):
-            flat = False
-            host = False
         else:
-            raise ValueError("unsupported input type")
+            raise ValueError(f"unsupported input type: {type(x).__name__}")
```
alexfikl, line 515: This is obviously unrelated. It just rearranges the order of the if statements so that the common `DOFArray` case is on top.

inducer:
LGTM, thanks!

PR #105, opened 2021-07-12
Allow subclasses of QBXLayerPotentialSource in copy

alexfikl:
I added some custom primitives to my code and had to subclass the various mappers to support them. 

For this PR, had to subclass `QBXLayerPotentialSource` to use a custom `QBXPreprocessor` in `preprocess_optemplate` (it just has a different `IdentityMapper` as a base).

There are a couple of places that also used some mappers with no easy way to replace them:
* `BoundExpression` uses the `EvaluationMapper` class in `eval`.
* `OperatorCompiler` uses the `OperatorCollector` in `__call__`.
* `OperatorCompiler` and `Code` use the `DependencyMapper` in a couple of places.

This `copy` was a bit too complex to copy paste, but the rest of them are not, so I'm not sure to what extent it's worth adding more arguments / attributes to make that easier.

### DIFF
--- a/pytential/qbx/__init__.py
+++ b/pytential/qbx/__init__.py
@@ -281,7 +281,7 @@ class QBXLayerPotentialSource(LayerPotentialSourceBase):
 
         # FIXME Could/should share wrangler and geometry kernels
         # if no relevant changes have been made.
-        return QBXLayerPotentialSource(
+        return type(self)(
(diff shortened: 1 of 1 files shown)


inducer:
This LGTM. I don't think I had really considered the "subclassed `QBXLayerPotentialSource`" use case, but I'm happy to consider changes that make that more usable.

PR #106, opened 2021-07-16
Improve GeometryCollection docs

alexfikl:
Was reading through the docs after inducer/grudge#122 and noticed they're a bit slim / outdated in places (pun intended).

### DIFF
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -634,39 +634,62 @@ class _GeometryCollectionConnectionCacheKey:
 class GeometryCollection:
     """A mapping from symbolic identifiers ("place IDs", typically strings)
     to 'geometries', where a geometry can be a
-    :class:`pytential.source.PotentialSource`
-    or a :class:`pytential.target.TargetBase`.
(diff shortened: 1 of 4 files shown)


pytential/symbolic/execution.py
```diff
@@ -705,6 +728,15 @@ def __init__(self, places, auto_where=None):
...
         # {{{ validate
 
+        # check auto_where
+        if auto_source.geometry not in self.places:
+            raise ValueError("'auto_where' source geometry is not in the "
+                f"collection: '{auto_source.geometry}'")
+
+        if auto_target.geometry not in self.places:
+            raise ValueError("'auto_where' target geometry is not in the "
+                f"collection: '{auto_target.geometry}'")
```
alexfikl, line 738: Is this too strict? It caused a few errors in the tests.

My reasoning was that this would fail later (if autotagging in `bind`) in a more cryptic way.

inducer: I agree that this check is a good idea.

inducer:
Thanks!

PR #109, opened 2021-08-17
Fix typo in test_layer_pot_identity

alexfikl:
'green_grad' was renamed to 'grad_green' at some point.

Also changed the exceptions to warnings, so the tests can still be run from the commandline.


inducer:
Thanks!

PR #114, opened 2021-09-26
Remove "where" deprecation warnings in primitives

alexfikl:
They were added in 2019 (unfortunately with no due date).


inducer:
Thanks! Let's get rid of it.

PR #118, opened 2021-10-22
Remove usage of meshmode.dof_array.flatten and friends

alexfikl:
Depends on
- [x] inducer/sumpy#89
- [x] ~~inducer/arraycontext#114~~
- [x] inducer/arraycontext#115
- [x] inducer/arraycontext#142


pytential/linalg/proxy.py
```diff
@@ -330,9 +329,9 @@ def __call__(self,
 
         # {{{ get proxy centers and radii
 
-        from arraycontext import thaw
-        from meshmode.dof_array import flatten
-        sources = flatten(thaw(discr.nodes(), actx))
+        sources = actx.np.reshape(
+                flatten(discr.nodes(), actx),
+                (self.ambient_dim, -1))
```
inducer, line 334: This is not an improvement IMO. The prior approach involved no copying, whereas this forces a copy. Do we have a way in `arraycontext` to flatten only below a certain `leaf_class`? If not, then I think this would be a good use case for it.

alexfikl: No, it just flattens the whole thing at the moment.

Is it worth adding there or should we just un-deprecate `meshmode.dof_array.flatten`? And implement it in terms of `arraycontext.flatten` mostly.

inducer: I think it makes sense to add to `arraycontext.flatten`. It might be relatively simple to do.

alexfikl: Assuming inducer/arraycontext#142 gets in, cce2995 removes all uses of `reshape + flatten` like that.

alexfikl: f9d5fcd should fix any remaining issues.

inducer:
LGTM, thanks!

PR #119, opened 2021-11-09
Use dataclasses instead of Record in compiler

alexfikl:
Was looking through this to get an idea of how many layer potentials / FMMs I have and cleaned it up a bit:
* used dataclasses and made attributes explicit in all the `Instruction` classes
* added some more type hints here and there


inducer:
LGTM, thanks!

PR #120, opened 2021-11-14
Fix _small_mat primitive helpers

alexfikl:
* `_small_mat_inverse` wasn't returning a matrix of the same size as the input in either case, so it couldn't do `inv @ x` afterwards.
* `_small_mat_eivenvalues` was written out in gory detail, making it hard to read; should be a bit clearer now (?)


inducer:
Thx!

PR #121, opened 2021-11-15
Optimize DerivativeBinder

alexfikl:
Actually uses inducer/pymbolic#72 to hopefully make `DerivativeBinder` a bit faster. 

These changes got the example in `test_derivative_binder_expr` down from 9-10s to 4-5s.
EDIT: Currently [2022-02-15] runs in ~3.6s on my machine.

* [x] Needs inducer/pymbolic#72.

### DIFF
--- a/pytential/symbolic/mappers.py
+++ b/pytential/symbolic/mappers.py
@@ -31,7 +31,6 @@ from pymbolic.mapper import (
         )
 from pymbolic.mapper.dependency import (
         DependencyMapper as DependencyMapperBase)
-from pymbolic.geometric_algebra import componentwise
 from pymbolic.geometric_algebra.mapper import (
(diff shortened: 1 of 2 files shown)


pytential/symbolic/mappers.py
```diff
@@ -160,13 +192,22 @@ def map_variable(self, expr):
...
+        if aggregate is expr.aggregate and index is expr.index:
+            return expr
+
+        return aggregate[index]
 
     map_q_weight = map_variable
     map_ones = map_variable
 
     def map_node_sum(self, expr):
-        return componentwise(type(expr), self.rec(expr.operand))
```
alexfikl, line 169: This `componentwise` is not needed because `NodeSum` already does it in its `__new__`

inducer:
LGTM, thanks!

PR #122, opened 2021-11-15
Add more types to OperatorCompiler

alexfikl:
* Adds more types.
* Changes some `IdentityMapper.__init__` to `super()`.
* Some f-strings here and there.

### DIFF
--- a/pytential/symbolic/compiler.py
+++ b/pytential/symbolic/compiler.py
@@ -22,15 +22,15 @@ THE SOFTWARE.
 
 from dataclasses import dataclass
 from functools import reduce
-from typing import Any, Callable, Dict, List, Optional, Set
+from typing import Any, Callable, Dict, Hashable, List, Optional, Set
 
 import numpy as np
 
-from pymbolic.primitives import cse_scope, Expression
+from pymbolic.primitives import cse_scope, Expression, Variable
 from pytools import memoize_method
 from sumpy.kernel import Kernel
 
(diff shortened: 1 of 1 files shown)


pytential/symbolic/compiler.py
```diff
@@ -465,44 +464,46 @@ def execute(self, exec_mapper, pre_assign_check=None):
...
+        return DependencyMapper(
+                # include_operator_bindings=False,
                 include_lookups=False,
                 include_subscripts=include_subscripts,
                 include_calls="descend_args")
 
-        return self.dep_mapper
+    @property
+    def dep_mapper(self):
+        return self.dep_mapper_factory()
```
alexfikl, line 499: This is probably a behavior change, but I didn't really understand what the intention of the original code was.

It gets memoized, but it also sets that `dep_mapper` attribute based on `include_subcripts` (?)

inducer: I agree that the old code looks might fishy.

inducer:
I was curious what it would take to check these types, which led to #123. (Answer: quite a bit!)

inducer:
Thanks, LGTM!

alexfikl:
> I was curious what it would take to check these types, which led to #123. (Answer: quite a bit!)

Yeah, I imagine it would take a bit of work to get anything useful, since `pymbolic` isn't typed either :(.

PR #124, opened 2021-12-11
Investigate stretch factors

alexfikl:
Added a simple case with a warped mesh and just a few elements to see what's going on and they don't seem to be discontinuous here. Not sure what it doesn't like about the torus.

On the other hand, the triangle and square cases seem to be seeing very different values.


alexfikl:
<p align="center">

</p>

Left is simplex and right is the tensor product. They're both nice and continuous, but annoyingly different.

alexfikl:
For future reference, a few remarks for this
* the simplex and quad cases have no right to match, since the simples is skewed to an equilateral reference element. That means that none of the vertex angles and edge lengths of the reference elements match between the two, so the stretch factors won't match either.
* stretch factors don't have a right to be continuous, since two adjacent elements can be stretched the same along one axis, but very differently along another.
* on the `gmsh` meshes, there's no difference (to eye ball norms) between simplices and quads.

PR #130, opened 2022-01-28
Introduce a small_sym_mat_eigenvalues

alexfikl:
This introduces and uses a `_small_sym_mat_eigenvalues` to remove the posibility of having a negative number in the square root when solving for the eigenvalues.

### DIFF
--- a/pytential/symbolic/primitives.py
+++ b/pytential/symbolic/primitives.py
@@ -855,10 +855,38 @@ def _small_mat_eigenvalues(mat):
     elif m == 2:
         (a, b), (c, d) = mat
         tr_mat = cse(a + d)
-        det_mat = a * d - b * c
 
         # solutions to lambda**2 - tr(A) * lambda + det(A)
-        sqrt_discriminant = cse(sqrt(tr_mat**2 - 4*det_mat))
+        # NOTE: 4 * b * c + (a - d)**2 can still become negative if the matrix
+        # is not positive definite, but there's not much we can do here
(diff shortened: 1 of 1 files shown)


pytential/symbolic/primitives.py
```diff
@@ -858,7 +858,7 @@ def _small_mat_eigenvalues(mat):
         det_mat = a * d - b * c
 
         # solutions to lambda**2 - tr(A) * lambda + det(A)
-        sqrt_discriminant = cse(sqrt(tr_mat**2 - 4*det_mat))
+        sqrt_discriminant = cse(sqrt(abs(tr_mat**2 - 4*det_mat)))
```
inducer, line 861: - Is it better to `abs` or to saturate at 0? (My gut feeling is saturate.)

inducer: Or maybe specialize to a symmetric eigensolve?

alexfikl: Went with specializing for a symmetric matrix, in which case that whole thing can be written as a sum of squares and is always nicely positive.

I also checked where it was failing, and there were just 3 nodes that had a value of `-4.4e-16`, so definitely some floating point issues.

inducer: Thanks for checking! That makes me less worried about this whole affair. And the symmetric eigenvalue thing seems like a reasonably clean and reliable way out of this.

inducer:
LGTM. Thanks!

PR #131, opened 2022-01-28
Misc ProxyGenerator changes

alexfikl:
This just takes out some small changes to the proxy parts from #30. The commits should be fairly self-contained and self-explanatory.


inducer:
LGTM, thanks!

PR #132, opened 2022-02-06
Fix scipy intersphinx link

alexfikl:
Seems to have moved (?) but couldn't find anything official, so not sure if this is just a temporary redirect because the main site is down for some reason.

https://docs.scipy.org/doc/scipy/reference

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -19,7 +19,7 @@ intersphinx_mapping = {
     "https://docs.python.org/3/": None,
     "https://documen.tician.de/boxtree/": None,
     "https://numpy.org/doc/stable/": None,
-    "https://docs.scipy.org/doc/scipy/reference/": None,
+    "https://scipy.github.io/devdocs/": None,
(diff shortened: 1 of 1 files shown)


inducer:
Thanks! I think this is just an oversight though. The `/devdocs` thing looks like it might be an in-development version of the docs, not sure that's a good thing to link to.

https://github.com/scipy/scipy.org/issues/452

should get us the answer we're looking for.

alexfikl:
> Thanks! I think this is just an oversight though. The `/devdocs` thing looks like it might be an in-development version of the docs, not sure that's a good thing to link to.
> 
> [scipy/scipy.org#452](https://github.com/scipy/scipy.org/issues/452)
> 
> should get us the answer we're looking for.

Thanks for the issue! I used that one because the `Documentation` tab on http://scipy.org seems to link to it, but yeah, `dev` does make it sound like it's not the stable version.

inducer:
Fair enough! It's easy to change back, so let's go with this.

PR #129, opened 2022-01-21
Per-node stretch factors

alexfikl:
Switched `_quad_resolution` to be per-node instead of per-element. Let's see what breaks!

~~@inducer The first commit is just #46 squashed down.~~

PR #135, opened 2022-02-19
Use element instead of panel nomenclature

alexfikl:
Matches the nomenclature used in meshmode and everywhere else!


alexfikl:
This should go in after #129, so I'm leaving it on draft for now.

inducer:
Well, this was trivial to review, so in it goes. :)

inducer:
Thanks for making this happen!

PR #139, opened 2022-04-01
IntG.copy: fix checks for new values

alexfikl:
I hit a bug (?) here due to
```
kernel_arguments = kernel_arguments or self.kernel_arguments
```
if `kernel_arguments = {}`, i.e. it wouldn't remove the arguments. This looks like a clearer way to check. What do you think?


inducer:
LGTM. Thanks!

PR #140, opened 2022-04-09
Add cache-preserving collection merging

alexfikl:
This adds a very dumb version of collection merging while keeping some caches.
* it just accepts points (i.e. `PointsTarget`), which should be harmless to add.
* it just copies over a few known caches.

How much smarter do you think we need to make this? The test does more or less what we would need from it for the proxy evaluation: adds another geometry in there, evaluates some layer potentials / expansions, dumps everything at the end.

### DIFF
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -23,7 +23,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import Optional
+from typing import Any, Dict, Hashable, Optional
 
 from pymbolic.mapper.evaluator import (
         EvaluationMapper as PymbolicEvaluationMapper)
@@ -273,9 +273,13 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
         else:
             return self.rec(expr.child)
 
+        # NOTE: the expr.prefix is added to the key mainly for testing purposes
+        # (i.e. to check if a given CSE is there in some simple cases)
+        key = (expr.prefix, expr.child)
+
         from numbers import Number
         try:
-            rec = cache[expr.child]
+            rec = cache[key]
             if (expr.scope == sym.cse_scope.DISCRETIZATION
                     and not isinstance(rec, Number)):
                 rec = thaw(rec, self.array_context)
@@ -285,7 +289,7 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
                     and not isinstance(rec, Number)):
                 cached_rec = freeze(cached_rec, self.array_context)
 
-            cache[expr.child] = cached_rec
+            cache[key] = cached_rec
 
         return rec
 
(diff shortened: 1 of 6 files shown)


pytential/symbolic/execution.py
```diff
@@ -711,28 +712,31 @@ def __init__(self, places, auto_where=None):
...
-            self.places[auto_source.geometry] = source_discr
-            self.places[auto_target.geometry] = target_discr
+            places_dict[auto_source.geometry] = source_discr
+            places_dict[auto_target.geometry] = target_discr
         else:
-            self.places = places
+            places_dict = places
 
+        import immutables
+        self.places = immutables.Map(places_dict)
```
alexfikl, line 735: Went ahead and made this immutable, just to be on the safe side. Should this use `pyrsistent` instead? I think `sumpy` depends on that already, right?

pytential/symbolic/execution.py
```diff
@@ -932,6 +934,37 @@ def __repr__(self):
...
 
+
+_KNOWN_GEOMETRY_COLLECTION_CACHE_KEYS = (
+        EvaluationMapperCSECacheKey,
+        _GeometryCollectionConnectionCacheKey,
+        _GeometryCollectionDiscretizationCacheKey,
+        )
+
+
+def add_geometry_collection(
```
inducer, line 945:
```suggestion
def add_geometry_to_collection(
```

inducer, line 945: Add to docs.

inducer, line 945: Add docs.

alexfikl: Renamed and added docs in https://github.com/inducer/pytential/compare/9d5eddbb52bc99495736f6349a9531e1463af545..1a41e0ba2726a623d153084ade2fbd20688cf07a.

alexfikl: Reworded that a bit in https://github.com/inducer/pytential/compare/547c7d34855dff898f6fba4f4ea4899a0ae3d708..d623df0e16902c87f21c7bc05a7967db5a3e0836 (probably best to just check d623df0)

pytential/symbolic/execution.py
```diff
@@ -274,8 +274,9 @@ def map_common_subexpression(self, expr):
             return self.rec(expr.child)
 
         from numbers import Number
+        key = (expr.prefix, expr.child)
```
inducer, line 277: How come the prefix became part of the key? (Add a comment to explain?)

pytential/symbolic/execution.py
```diff
@@ -932,6 +934,51 @@ def __repr__(self):
...
+
+        if not isinstance(geometry, (PointsTarget, PointPotentialSource)):
+            raise TypeError(
+                    f"Cannot add a geometry of type '{type(geometry).__name__}' "
+                    "to the existing collection. Construct a new collection "
+                    "instead.")
+
+    # copy over all the caches
+    new_places = places.merge(geometries)
+    for key in _KNOWN_GEOMETRY_COLLECTION_CACHE_KEYS:
```
inducer, line 976: Why not iterate over all existing cache names and at least warn if one is encountered that's not understood?

inducer (commented):
Thanks! Two additional questions upon a more careful read.

pytential/symbolic/execution.py
alexfikl: Ah, I'll remove that. It was initially added in the hopes of doing something more intelligent in the geometry merging (e.g. only copy over the expressions that have `DOFDescriptors` that are still in the collection)

alexfikl: Added comment in https://github.com/inducer/pytential/compare/d623df0e16902c87f21c7bc05a7967db5a3e0836..5ba6baee5fa0bded87ab97ed9fc173d6b846aee2.

inducer:
That looks good. Thanks!

PR #143, opened 2022-04-15
Proxy: reuse TreeBuilder 

alexfikl:
Main changes are
* use `QBXLayerPotentialSource.tree_code_container.build_tree` if available when gathering neighbors and partitioning things.
* add a `build_area_query()` to `tree_code_container` and reuse that as well.

The rest of it is just moving some things around to make sure that everyone has proper access to the geometry collection and the `QBXLayerPotentialSource` when needed.

### DIFF
--- a/pytential/linalg/proxy.py
+++ b/pytential/linalg/proxy.py
@@ -21,17 +21,16 @@ THE SOFTWARE.
 """
 
 from dataclasses import dataclass
-from typing import Callable, Optional
+from typing import Any, Callable, Optional
 
 import numpy as np
 import numpy.linalg as la
 
 from arraycontext import PyOpenCLArrayContext, flatten
-from meshmode.discretization import Discretization
 from meshmode.dof_array import DOFArray
 
 from pytools import memoize_in
-from pytential.qbx import QBXLayerPotentialSource
+from pytential import GeometryCollection, bind, sym
 from pytential.linalg.utils import BlockIndexRanges
 
 import loopy as lp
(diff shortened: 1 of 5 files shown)


pytential/linalg/proxy.py
```diff
@@ -68,15 +71,23 @@ def partition_by_nodes(
...
+    discr = places.get_discretization(dofdesc.geometry, dofdesc.discr_stage)
 
-        builder = TreeBuilder(actx.context)
+    if tree_kind is not None:
+        from pytential.qbx import QBXLayerPotentialSource
+        if isinstance(lpot_source, QBXLayerPotentialSource):
+            builder = lpot_source.tree_code_container.build_tree()
+        else:
+            from boxtree import TreeBuilder
+            builder = TreeBuilder(actx.context)
```
inducer, line 90: Explain or remove re-create-y case.

inducer (approved):
:tada: Thanks!

pytential/linalg/proxy.py
alexfikl: @inducer Wanted to do this, but wasn't sure which array context: the one the user has or the `_setup_actx` of the `QBXLayerPotentialSource`?

alexfikl: In the meantime, the latest rebase removed the `if` since nobody was using the fallback.

This is mostly good to go on my end and we can move the several code containers that are now attached to `QBXLayerPotentialSource` in a separate PR?

inducer: #145

inducer:
LGTM. Thanks!

PR #146, opened 2022-04-18
Add a pytential specific DefaultExpansionFactory to get a QBX local expansion

isuruf:
A use-case is to use a FFT based M2L and non-FFT based M2QBXL.


inducer:
Thanks!

PR #147, opened 2022-04-18
Make code containers functions instead of methods

### LINKED ISSUE #145: Move code container retrieval to a function operating on the array context
x-ref: https://github.com/inducer/pytential/pull/143#discussion_r851572652

To answer the question posed there, my gut currently says that trees and such (and therefore their builders) belong in the setup array context, and should be "thawed into" the computation actx on use. @alexfikl, do you have a feeling on that?

alexfikl:
How does this look? It moves each code container from `QBXLayerPotentialSource`, e.g. `refiner_code_container` goes to `pytential.qbx.refinement` as a little function.

Fixes #145.

### DIFF
--- a/pytential/collection.py
+++ b/pytential/collection.py
@@ -274,12 +274,12 @@ class GeometryCollection:
         except KeyError:
             dofdesc = sym.DOFDescriptor(geometry, discr_stage)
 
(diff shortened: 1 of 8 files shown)


inducer:
This looks good. Thanks!

PR #152, opened 2022-05-08
Add mappers for direct solver

alexfikl:
This rips out the mappers from #30 for some discussion + tests. The new mappers are

* `IntGTermCollector`: removes all non-`IntG` terms in the expression and all non-constant coefficients.
* `KernelTransformationRemover`: removes transforms from `IntG` kernels.
* `DOFDescriptorReplacer`: re-tags an expression.

They're basically used to clean up a given expression for source->proxy or proxy->target evaluation, which 
* doesn't need any non-`IntG` terms, since those are only diagonal operators.
* doesn't need to see any source / target derivatives, but generally also can't use these because we don't want to compute normals and things at the proxy points.

TODO:
* [x] Requires inducer/pymbolic#102.

### DIFF
--- a/pytential/symbolic/mappers.py
+++ b/pytential/symbolic/mappers.py
@@ -68,6 +68,8 @@ def rec_int_g_arguments(mapper, expr):
     return densities, kernel_arguments, changed
 
 
+# {{{ IdentityMapper
+
 class IdentityMapper(IdentityMapperBase):
     def map_node_sum(self, expr):
         operand = self.rec(expr.operand)
@@ -137,6 +139,10 @@ class IdentityMapper(IdentityMapperBase):
 
         return type(expr)(expr.from_dd, expr.to_dd, operand)
 
+# }}}
+
+
+# {{{ CombineMapper
 
 class CombineMapper(CombineMapperBase):
     def map_node_sum(self, expr):
@@ -169,6 +175,10 @@ class CombineMapper(CombineMapperBase):
 
     map_error_expression = map_is_shape_class
 
+# }}}
+
+
+# {{{ Collector
 
 class Collector(CollectorBase, CombineMapper):
     def map_ones(self, expr):
@@ -187,6 +197,10 @@ class OperatorCollector(Collector):
 class DependencyMapper(DependencyMapperBase, Collector):
     pass
 
+# }}}
+
+
+# {{{ EvaluationMapper
 
 class EvaluationMapper(EvaluationMapperBase):
     """Unlike :mod:`pymbolic.mapper.evaluation.EvaluationMapper`, this class
@@ -250,8 +264,10 @@ class EvaluationMapper(EvaluationMapperBase):
                 expr.prefix,
                 expr.scope)
 
+# }}}
 
-# {{{ dofdesc tagging
+
+# {{{ LocationTagger
 
 class LocationTagger(CSECachingMapperMixin, IdentityMapper):
     """Used internally by :class:`ToTargetTagger`."""
@@ -384,6 +400,10 @@ class ToTargetTagger(LocationTagger):
         self.operand_rec = LocationTagger(default_source,
                                           default_source=default_source)
 
+# }}}
+
+
+# {{{ DiscretizationStageTagger
 
 class DiscretizationStageTagger(IdentityMapper):
     """Descends into an expression tree and changes the
@@ -427,7 +447,7 @@ class DiscretizationStageTagger(IdentityMapper):
 # }}}
 
 
-# {{{ derivative binder
+# {{{ DerivativeBinder
 
 class DerivativeTaker(Mapper):
     def __init__(self, ambient_axis):
@@ -498,7 +518,7 @@ class DerivativeBinder(DerivativeBinderBase, IdentityMapper):
 # }}}
 
 
-# {{{ Unregularized preprocessor
+# {{{ UnregularizedPreprocessor
 
 class UnregularizedPreprocessor(IdentityMapper):
 
@@ -524,7 +544,7 @@ class UnregularizedPreprocessor(IdentityMapper):
 # }}}
 
 
-# {{{ interpolation preprocessor
+# {{{ InterpolationPreprocessor
 
 class InterpolationPreprocessor(IdentityMapper):
     """Handle expressions that require upsampling or downsampling by inserting
@@ -596,7 +616,7 @@ class InterpolationPreprocessor(IdentityMapper):
 # }}}
 
 
-# {{{ QBX preprocessor
+# {{{ QBXPreprocessor
 
 class QBXPreprocessor(IdentityMapper):
     def __init__(self, geometry, places):
@@ -656,7 +676,7 @@ class QBXPreprocessor(IdentityMapper):
 # }}}
 
 
-# {{{ stringifier
+# {{{ StringifyMapper
 
 def stringify_where(where):
     return str(prim.as_dofdesc(where))
@@ -769,12 +789,12 @@ class StringifyMapper(BaseStringifyMapper):
         return "IsShape[{}]({})".format(stringify_where(expr.dofdesc),
                                         expr.shape.__name__)
 
-# }}}
-
 
 class PrettyStringifyMapper(
         CSESplittingStringifyMapperMixin, StringifyMapper):
     pass
+
+# }}}
 
 
 # {{{ graphviz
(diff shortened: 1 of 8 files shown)


pytential/symbolic/mappers.py
```diff
@@ -187,6 +232,45 @@ class DependencyMapper(DependencyMapperBase, Collector):
     pass
 
 
+class DOFDescriptorCollector(Collector):
```
alexfikl, line 235: This one is just added for testing. Should it just be in `test_symbolic.py`? Not just if it's more generally useful..

pytential/symbolic/pde/scalar.py
```diff
@@ -400,8 +400,10 @@ def Sp(density):
             if isinstance(self.kernel, HelmholtzKernel):
                 DpS0u = (
                         sym.Dp(
-                            self.kernel - self.laplace_kernel,
-                            laplace_s_inv_sqrt_w_u,
+                            self.kernel, laplace_s_inv_sqrt_w_u,
+                            qbx_forced_limit=+1, **kwargs)
+                        - sym.Dp(
+                            self.laplace_kernel, laplace_s_inv_sqrt_w_u,
```
alexfikl, line 406: This seems to have been broken for quite a while!

pytential/symbolic/mappers.py
```diff
@@ -137,6 +139,45 @@ def map_interpolation(self, expr):
 
         return type(expr)(expr.from_dd, expr.to_dd, operand)
 
+# }}}
+
+
+# {{{ IntGTermCollector
+
+class IntGTermCollector(IdentityMapper):
```
inducer, line 147: Add docstrings, add to docs?

pytential/symbolic/mappers.py
```diff
@@ -384,6 +512,46 @@ def __init__(self, default_source, default_target):
                                           default_source=default_source)
 
 
+class LocationReplacer(LocationTagger):
```
inducer, line 515: Add docstring describing purpose?

inducer, line 515: Maybe underscore?

pytential/symbolic/mappers.py
```diff
@@ -384,6 +512,46 @@ def __init__(self, default_source, default_target):
...
+                    name: self.operand_rec(arg_expr)
+                    for name, arg_expr in expr.kernel_arguments.items()
+                    }
+                )
+
+
+class DOFDescriptorReplacer(LocationReplacer):
+    r"""Mapper that replaces all the
+    :class:`~pytential.symbolic.primitives.DOFDescriptor`\ s in the expression
+    with the given ones.
```
inducer, line 535: Maybe add use case to docstring?

pytential/symbolic/pde/scalar.py
inducer: You mean the difference kernels no longer work? Yikes. Should have had tests for them. I guess it makes sense that FMMs for them don't work directly (no PDE after all), but for the P2P, there's no real reason to not support them.

@isuruf What do you think?

Also, this is maybe a separate discussion from this PR.

pytential/symbolic/mappers.py
```diff
@@ -249,8 +333,52 @@ def map_common_subexpression(self, expr):
...
+# {{{ KernelTransformationRemover
+
+class KernelTransformationRemover(IdentityMapper):
+    r"""A mapper that removes the transformations from the kernel of all
+    :class:`~pytential.symbolic.primitives.IntG`\ s in the expression.
+
+    This includes source and target derivatives and other such transformations.
+    Any unnecessary kernel arguments are also removed from
+    :class:`~pytential.symbolic.primitive.IntG.kernel_arguments`.
+    """
```
inducer, line 348: Describe example use case in docstring?

pytential/symbolic/mappers.py
inducer: I'm not sure I see the general use case, so maybe yeah, my vote would be to move to `testlib` or so.

inducer (commented):
Thanks! Some comments from a quick scroll.

pytential/symbolic/mappers.py
alexfikl: Got it, moved in the latest version!

pytential/symbolic/mappers.py
alexfikl: Added in dfcee25. Let me know what you think!

The mappers aren't actually hooked into the docs, so I didn't add them anywhere. Should they be?

pytential/symbolic/mappers.py
alexfikl: Added in dfcee25.

alexfikl: Done in dfcee25.

pytential/symbolic/mappers.py
alexfikl: Added a little blurb in dfcee25. Let me know what you think.

pytential/symbolic/pde/scalar.py
alexfikl: Yes, I meant the difference between `sumpy` kernels in there. I was the last one who touched this code, but it seems to have been this way for a while, so not sure when it broke.

Agreed this is for a different PR, just wanted to make a note of it before I forgot.

pytential/symbolic/mappers.py
alexfikl: Added something in aab3e5b. Let me know what you think.

pytential/symbolic/mappers.py
```diff
@@ -137,6 +139,66 @@ def map_interpolation(self, expr):
 
         return type(expr)(expr.from_dd, expr.to_dd, operand)
 
+# }}}
+
+
+# {{{ IntGTermCollector
+
+class IntGTermCollector(IdentityMapper):
```
inducer, line 147: Having these mappers as part of the global `mappers` module is confusing to my mind. Maybe move them someplace more direct-solver-specific? If a more generic use case shows up, they can always be fished out.

pytential/symbolic/mappers.py
inducer: Once they're in a module of direct-solver-specific mappers (see below) (that's perhaps hooked into the docs in a place marked "internal API only"), I don't see why not.

pytential/symbolic/mappers.py
```diff
@@ -137,6 +139,66 @@ def map_interpolation(self, expr):
...
+    def map_product(self, expr):
+
+        collector = OperatorCollector()
+
+        from pymbolic.primitives import is_constant
+        children = []
+        for child in expr.children:
+            rec_child = self.rec(child)
+            if is_constant(child) or collector(rec_child):
+                children.append(rec_child)
```
inducer, line 190: Maybe complain if there are more than one of these? Otherwise this is unlikely to be a linear operator...

pytential/symbolic/mappers.py
inducer: LGTM!

pytential/symbolic/pde/scalar.py
inducer: #156

test/test_symbolic.py
```diff
@@ -394,6 +394,98 @@ def test_derivative_binder_expr():
...
+@pytest.mark.parametrize("k", [0, 5])
+def test_mapper_int_g_term_collector(op_name, k):
+    ambient_dim = 3
+    op = _make_operator(ambient_dim, op_name, k)
+    expr = op.operator(op.get_density_var("sigma"))
+
+    from pytential.symbolic.mappers import IntGTermCollector
+    expr_only_intgs = IntGTermCollector()(expr)
+
+    # FIXME: how to check this did something?
```
inducer, line 457: Hardcode an expected result?

inducer (commented):
Thanks! A few more things below. LGTM once these are addressed.

inducer:
Unsubscribing... @-mention or request review once it's ready for a look or needs attention.

pytential/symbolic/mappers.py
alexfikl: Moved to `pytential.linalg.direct_solver_symbolic` in latest version

pytential/symbolic/mappers.py
alexfikl: Yep, done in d124100.

test/test_symbolic.py
alexfikl: Yeah, did that in 3ef6f4f.

pytential/symbolic/mappers.py
alexfikl: Yeah, good point! Changed in cf1d364.

inducer:
This looks good, thanks!

PR #153, opened 2022-05-09
Rename nterms to order for boxtree compatibility

gaohao95:
- [x] Mutually dependent on https://github.com/inducer/pytential/pull/153


inducer:
Thanks! On its way in.

PR #160, opened 2022-06-11
Port deprecated uses of freeze and thaw

alexfikl:
Last one!


inducer:
Wow, you're on a roll. :) Thanks!

PR #167, opened 2022-07-04
Investigate FMM assertion

alexfikl:
This seems to be failing locally and wanted to see if it's something weird in my setup.

### DIFF
--- a/test/test_layer_pot.py
+++ b/test/test_layer_pot.py
@@ -612,6 +612,50 @@ def test_3d_jump_relations(actx_factory, relation, visualize=False):
 # }}}
 
 
+# {{{
+
+def test_qbx_fmm_failure(actx_factory, visualize=False):
+    actx = actx_factory()
+
+    # {{{ setup
+
+    from extra_int_eq_data import CircleTestCase
+    case = CircleTestCase(
+        target_order=5,
+        qbx_order=5,
+        source_ovsmp=4,
+        fmm_order=10, fmm_backend="sumpy",
+        )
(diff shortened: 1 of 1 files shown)


alexfikl:
For some details, this fails on this check in `sumpy`
https://github.com/inducer/sumpy/blob/main/sumpy/fmm.py#L649
where `pot` is just list of 13 empty `cl.Array` for some reason. The 13 part comes from `tree_indep.target_kernels` having 13 elements for this test case. The reason they're empty is because `output_zeros` in 

https://github.com/inducer/pytential/blob/2d95be0a41c70cf90a3cf398d501d83427270f96/pytential/qbx/fmm.py#L135

finds that `nqbtl.nfiltered_targets = 0`. From looking at the function

https://github.com/inducer/pytential/blob/2d95be0a41c70cf90a3cf398d501d83427270f96/pytential/qbx/geometry.py#L849-L864

it seems like it doesn't find any targets that need QBX, which is just weird..

alexfikl:
Simplified the original Beltrami test case to a simpler setup that fails in the same way. It just evaluates
```
sym.Spp(LaplaceKernel, sigma) + sym.Sp(LaplaceKernel, sigma)
```
and raises the same assertion error in `sumpy`.

inducer:
Looks like https://github.com/inducer/loopy/pull/667 addressed this issue.

alexfikl:
> Looks like [inducer/loopy#667](https://github.com/inducer/loopy/pull/667) addressed this issue.

Yep, it's fixed! Awesome!

PR #169, opened 2022-07-06
Simplify elementwise reductions

alexfikl:
Mostly switches to using a list comprehension, which seems cleaner.

### DIFF
--- a/pytential/symbolic/execution.py
+++ b/pytential/symbolic/execution.py
@@ -124,29 +124,31 @@ class EvaluationMapperBase(PymbolicEvaluationMapper):
     def _map_elementwise_reduction(self, reduction_name, expr):
         import loopy as lp
         from arraycontext import make_loopy_program
-        from meshmode.transform_metadata import (
-                ConcurrentElementInameTag, ConcurrentDOFInameTag)
+        from meshmode.transform_metadata import ConcurrentElementInameTag
+        actx = self.array_context
 
-        @memoize_in(self.places, "elementwise_node_"+reduction_name)
+        @memoize_in(actx, (
+            EvaluationMapperBase._map_elementwise_reduction,
+            f"elementwise_node_{reduction_name}"))
         def node_knl():
             t_unit = make_loopy_program(
                     """{[iel, idof, jdof]:
                         0<=iel<nelements and
                         0<=idof, jdof<ndofs}""",
                     """
-                    result[iel, idof] = %s(jdof, operand[iel, jdof])
+                    <> el_result = %s(jdof, operand[iel, jdof])
+                    result[iel, idof] = el_result
                     """ % reduction_name,
-                    name="nodewise_reduce")
+                    name=f"elementwise_node_{reduction_name}")
 
             return lp.tag_inames(t_unit, {
                 "iel": ConcurrentElementInameTag(),
-                "idof": ConcurrentDOFInameTag(),
                 })
 
-        @memoize_in(self.places, "elementwise_"+reduction_name)
+        @memoize_in(actx, (
+            EvaluationMapperBase._map_elementwise_reduction,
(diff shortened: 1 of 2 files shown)


pytential/symbolic/execution.py
```diff
@@ -127,27 +127,27 @@ def _map_elementwise_reduction(self, reduction_name, expr):
...
             return lp.tag_inames(t_unit, {
                 "iel": ConcurrentElementInameTag(),
                 "idof": ConcurrentDOFInameTag(),
                 })
 
-        @memoize_in(self.places, "elementwise_"+reduction_name)
+        @memoize_in(actx, f"elementwise_element_{reduction_name}")
         def element_knl():
-            # FIXME: This computes the reduction value redundantly for each
-            # output DOF.
```
alexfikl, line 150: I tried fixing that `FIXME` by storing the result of the elementwise reduction. Does that look better?

alexfikl: Doesn't seem like this should be `ConcurrentDOFInameTag` anymore for `idof`?

inducer: Doesn't look like it is?

alexfikl: The `idof` from the `elementwise_node` kernel does have that tag (on line 146).

pytential/symbolic/execution.py
```diff
@@ -126,27 +126,27 @@ def _map_elementwise_reduction(self, reduction_name, expr):
         from arraycontext import make_loopy_program
         from meshmode.transform_metadata import (
                 ConcurrentElementInameTag, ConcurrentDOFInameTag)
+        actx = self.array_context
 
-        @memoize_in(self.places, "elementwise_node_"+reduction_name)
+        @memoize_in(actx, f"elementwise_node_{reduction_name}")
```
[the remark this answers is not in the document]
alexfikl: Done in
https://github.com/inducer/pytential/compare/cff9f643c2d68f6ef7c46cd2e20fb32d674ef47a..1ab635b81b52e49a3c5fb933f9667e50fbc78f78

pytential/symbolic/execution.py
alexfikl: Apparently none of the tests were actually using this with `GRANULARITY_NODE`! I added one in https://github.com/inducer/pytential/compare/8ab424c713e31b03cd30864c2b479a8357263a83..fd3e217d830eda7cbb991c7278db07a6c3f05475.

This was failing with 
```
loopy.diagnostic.LoopyError: instruction 'insn_jdof_init' does not use all local hw axes (available: 0 used: ). Calling loopy.add_inames_for_unused_hw_axes(...) might help.
```
when `"idof": ConcurrentDOFInameTag()` was in the tags. Not sure if the best solution is to remove that, but it does feel like some sort of barrier / sync is needed to compute `el_result` before broadcasting it to all the `idofs`.

alexfikl:
@inducer I'm getting a failure on `Pytest Mac` a bit reliably for
```
worker 'gw1' crashed while running "test/test_layer_pot_eigenvalues.py::test_sphere_eigenvalues[<PyOpenCLArrayContext for <pyopencl.Device 'pthread' on 'Portable Computing Language'>>-sumpy-2-3-3]"
```
see https://github.com/inducer/pytential/runs/7625741382?check_suite_focus=true and https://github.com/inducer/pytential/runs/7625724520?check_suite_focus=true. Any idea what that could be about? Out of memory?

pytential/symbolic/execution.py
inducer: I pushed a revert of this before changing my mind. It's a bit of a CPU vs GPU conundrum, but since pytential is still a predominantly-CPU code, the lack of parallelism is not a tremendous concern.

inducer:
Thanks for working on this!

PR #170, opened 2022-07-07
Improve QBXLayerPotentialSource docs

alexfikl:
* adds some typing annotations (not super accurate, but better)
* adds some more docs for the experimental flags in there, since I was always confused where each one went.
* adds some `assert` for the various arguments
* removed `target_stick_out_factor` since the message said it should be gone by 2018!


pytential/qbx/__init__.py
```diff
@@ -87,111 +90,135 @@ class QBXLayerPotentialSource(LayerPotentialSourceBase):
...
 
         if _from_sep_smaller_crit is None:
             # This seems to win no matter what the box extent norm is
             # https://gitlab.tiker.net/papers/2017-qbx-fmm-3d/issues/10
             _from_sep_smaller_crit = "precise_linf"
+        assert isinstance(_from_sep_smaller_crit, str)
 
         if fmm_level_to_order is None:
-            if fmm_order is False:
+            if fmm_order is False or fmm_order is None:
```
alexfikl, line 199: Not sure what was meant to happen when both `fmm_level_to_order` and `fmm_order` are `None`?

pytential/qbx/__init__.py
```diff
@@ -87,111 +90,135 @@ class QBXLayerPotentialSource(LayerPotentialSourceBase):
...
             if density_discr.ambient_dim == 2:
                 # FIXME: This should be verified now that l^2 is the default.
                 _max_leaf_refine_weight = 64
             elif density_discr.ambient_dim == 3:
-                # For static_linf/linf: https://gitlab.tiker.net/papers/2017-qbx-fmm-3d/issues/8#note_25009  # noqa
-                # For static_l2/l2: https://gitlab.tiker.net/papers/2017-qbx-fmm-3d/issues/12  # noqa
+                # For static_linf/linf:
+                #   https://gitlab.tiker.net/papers/2017-qbx-fmm-3d/issues/8#note_25009
+                # For static_l2/l2:
+                #   https://gitlab.tiker.net/papers/2017-qbx-fmm-3d/issues/12
```
alexfikl, line 216: These links don't seem to be accessible from the outside?

pytential/qbx/__init__.py
inducer: Good question. I'd say error, we'll see how badly that blows up.

alexfikl: It did that before (because I forgot to check for `None`) and there are just 8 errors:
https://github.com/inducer/pytential/runs/7228738492?check_suite_focus=true

That's not so bad, so I'll just fix those up and raise an error.

pytential/qbx/__init__.py
alexfikl: That's fair! I'm not advocating to necessarily make them public or anything, so added a little note in https://github.com/inducer/pytential/compare/75d27feb90f28eab9ac559dd66c14cb25383f007..5349918da7ec88001f8a9f6dbf9fa69a3528747e

inducer:
LGTM, thanks!

PR #171, opened 2022-07-22
Move pytential.solve to pytential.linalg.gmres

alexfikl:
https://gitlab.tiker.net/inducer/pytential/-/issues/101


inducer:
Thanks!

PR #178, opened 2022-08-17
Cache the TranslationClassesBuilder

alexfikl:
Takes the construction from https://github.com/inducer/sumpy/blob/3d2a57aadc55c57c457fdb26ed1e28410ee2f54c/sumpy/fmm.py#L325-L331 and puts it in here for caching in the array context (like all the other builders / code containers around).

### DIFF
--- a/pytential/qbx/fmm.py
+++ b/pytential/qbx/fmm.py
@@ -112,13 +112,24 @@ non_qbx_box_target_lists`),
     def __init__(self, tree_indep, geo_data, dtype,
             qbx_order, fmm_level_to_order,
             source_extra_kwargs, kernel_extra_kwargs,
+            translation_classes_data=None,
             _use_target_specific_qbx=None):
         if _use_target_specific_qbx:
             raise ValueError("TSQBX is not implemented in sumpy")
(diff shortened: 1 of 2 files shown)


pytential/qbx/fmm.py
```diff
@@ -112,13 +112,24 @@ class QBXExpansionWrangler(SumpyExpansionWrangler):
...
             raise ValueError("TSQBX is not implemented in sumpy")
 
+        base_kernel = tree_indep.get_base_kernel()
+        if translation_classes_data is None and base_kernel.is_translation_invariant:
+            from pytential.qbx.fmm import translation_classes_builder
+            traversal = geo_data.traversal()
+            actx = geo_data._setup_actx
+
+            translation_classes_data, _ = translation_classes_builder(actx)(
+                actx.queue, traversal, traversal.tree, is_translation_per_level=True)
```
inducer, line 127: Thanks for highlighting `is_translation_per_level=True`. I suspect we could do a finer-grained job here, but this will do for now.

inducer:
Thanks!

PR #179, opened 2022-09-06
Updated sumpy and boxtree to arraycontext branches

alexfikl:
This is completely empty at the moment until upstreams actually work.


doc/conf.py
```diff
@@ -22,6 +22,10 @@
         "DOFDescriptorLike": "pytential.symbolic.dof_desc.DOFDescriptorLike",
         }
 
+nitpick_ignore_regex = [
+        ["py:class", r"boxtree\.array_context\.(.+)"],
+        ]
```
alexfikl, line 27: Need to remove this once the boxtree changes go in.

pytential/array_context.py
```diff
@@ -0,0 +1,89 @@
...
+        register_pytest_array_context_factory)
+
+__doc__ = """
+.. autoclass:: PyOpenCLArrayContext
+"""
+
+
+# {{{ PyOpenCLArrayContext
+
+class PyOpenCLArrayContext(SumpyPyOpenCLArrayContext):
```
alexfikl, line 40: How should this inherit from the `meshmode` and the `sumpy` array contexts? Presumably it needs functionality from both to handle the mesh and the FMM..

alexfikl:
@inducer Fixed this part too, so it seems like boxtree->sumpy->pytential all work with `arraycontext` now :grin:

PR #180, opened 2022-09-11
Add an assert that 2*something didn't change dtype in QBX tree

inducer:
This adds an assert to catch the specific pyopencl thing that is getting unbroken by https://github.com/inducer/pyopencl/pull/609.

- [x] Clearly, this is expected to fail until https://github.com/inducer/pyopencl/pull/609 is in.

### DIFF
--- a/pytential/qbx/utils.py
+++ b/pytential/qbx/utils.py
@@ -382,6 +382,9 @@ def build_tree_with_qbx_metadata(actx: PyOpenCLArrayContext,
             2 * qbx_element_to_source_starts
             if not use_stage2_discr
             else None)
+    if qbx_element_to_center_starts is not None:
+        assert (qbx_element_to_center_starts.dtype
+                == qbx_element_to_source_starts.dtype)
 
     # Transfer all tree attributes.
     tree_attrs = {}


test/test_global_qbx.py
```diff
@@ -257,21 +257,14 @@ def check_quad_res_to_helmholtz_k_ratio(element):
...
-        success = True
         for element_2 in iter_elements(quad_stage2_density_discr):
-            result = check_sufficient_quadrature_resolution(element_1, element_2)
-            success = success and result
-        assert success, "'check_sufficient_source_quadrature_resolution' failed"
+            assert check_sufficient_quadrature_resolution(element_1, element_2)
 
         if helmholtz_k is not None:
-            success = check_quad_res_to_helmholtz_k_ratio(element_1)
-            assert success, "'check_quad_res_to_helmholtz_k_ratio' failed"
```
inducer, line 274: These are equivalent, right? I have no idea why they were written in this contrived way.

pytential/qbx/utils.py
```diff
@@ -382,6 +382,8 @@ def _make_centers(discr):
             2 * qbx_element_to_source_starts
             if not use_stage2_discr
             else None)
+    if qbx_element_to_center_starts is not None:
+        assert qbx_element_to_center_starts.dtype == qbx_element_to_source_starts
```
alexfikl, line 386:
```suggestion
        assert qbx_element_to_center_starts.dtype == qbx_element_to_source_starts.dtype
```
?

alexfikl (approved):
Small typo, but otherwise this looks good to me! Thanks!

test/test_global_qbx.py
alexfikl: Hm, the original version would print out all the problematic elements before asserting, but otherwise they should be the same, yeah.

inducer: Hmm, OK. I think I'll drop this change then.

pytential/qbx/utils.py
inducer: That's weird. How did this ever pass then?

alexfikl: Ha.. no idea :confused:

inducer: Oh numpy. :facepalm: 

```python
>>> class A:
...    pass
... 
>>> a = A()
>>> a.dtype = np.dtype(np.int32)
>>> a.dtype == a
True
```

inducer:
Thanks for taking a look@

PR #181, opened 2022-09-12
Fix visualization in test_global_qbx

alexfikl:
Hit this while starting to look at inducer/pyopencl#609 yesterday.

The issue is that `lpot.density_discr` can be different to the stage1 discr, which caused an DOFArray inconsistency error when visualizing.


inducer:
Thx!

PR #185, opened 2022-11-11
Add pyproject and cython build dep

alexfikl:
Currently the `cython` dep doesn't seem to be declared anywhere (?).


inducer:
Thanks!

PR #188, opened 2022-12-05
Introduce an identifier for the default source/target for a collection

isuruf:
And rename DEFAULT_SOURCE -> UNNAMED_SOURCE

### DIFF
--- a/pytential/collection.py
+++ b/pytential/collection.py
@@ -142,8 +142,8 @@ class GeometryCollection:
             descriptors are used to define the default source and target
             geometries for layer potential evaluations.
             By default, they are set to
-            :class:`~pytential.symbolic.dof_desc.DEFAULT_SOURCE` and
-            :class:`~pytential.symbolic.dof_desc.DEFAULT_TARGET` for
+            :class:`~pytential.symbolic.dof_desc.UNNAMED_SOURCE` and
+            :class:`~pytential.symbolic.dof_desc.UNNAMED_TARGET` for
             sources and targets, respectively.
         """
 
--- a/pytential/collection.py
+++ b/pytential/collection.py
@@ -141,10 +141,7 @@ class GeometryCollection:
             :func:`~pytential.symbolic.dof_desc.as_dofdesc`. The two
             descriptors are used to define the default source and target
             geometries for layer potential evaluations.
-            By default, they are set to
-            :class:`~pytential.symbolic.dof_desc.UNNAMED_SOURCE` and
-            :class:`~pytential.symbolic.dof_desc.UNNAMED_TARGET` for
-            sources and targets, respectively.
+            By default, they are set to unspecified, but unique objects.
         """
 
         # {{{ construct dict
--- a/pytential/symbolic/dof_desc.py
+++ b/pytential/symbolic/dof_desc.py
@@ -71,13 +71,13 @@ class UNNAMED_TARGET:                   # noqa: N801
 class DEFAULT_SOURCE:                   # noqa: N801
     """Symbolic identifier for the default source. Geometries with
     this value get replaced with the default source given to
-    :func:`pytential.symbolic.execution.bind`."""
+    :func:`pytential.bind`."""
 
 
 class DEFAULT_TARGET:                   # noqa: N801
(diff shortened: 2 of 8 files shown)


alexfikl:
This looks good to me, but could use a bit more documentation. Some things that come to mind:
* Why can't we use `None` for `UNNAMED_SOURCE` and `UNNAMED_TARGET`?
* `DEFAULT_SOURCE` and `DEFAULT_TARGET` are replaced in `LocationTagger` with whatever it was given in the constructor, so not necessarily the `auto_source` and `auto_target` of the `GeometryCollection`. This disagrees with their docs?
* Are `DEFAULT_SOURCE` and `DEFAULT_TARGET` supposed to be for internal use now? i.e. stuff in `pytential.symbolic`.

isuruf:
> Why can't we use None for UNNAMED_SOURCE and UNNAMED_TARGET?

They need to be distinct values in order to be placed in a dictionary.

> DEFAULT_SOURCE and DEFAULT_TARGET are replaced in LocationTagger with whatever it was given in the constructor, so not necessarily the auto_source and auto_target of the GeometryCollection. This disagrees with their docs?

Fixed now.

> Are DEFAULT_SOURCE and DEFAULT_TARGET supposed to be for internal use now? i.e. stuff in pytential.symbolic.

Nope. UNNAMED_SOURCE and UNNAMED_TARGET are internal use only. Added to docs.

pytential/collection.py
```diff
@@ -142,8 +142,8 @@ def __init__(self,
             descriptors are used to define the default source and target
             geometries for layer potential evaluations.
             By default, they are set to
-            :class:`~pytential.symbolic.dof_desc.DEFAULT_SOURCE` and
-            :class:`~pytential.symbolic.dof_desc.DEFAULT_TARGET` for
+            :class:`~pytential.symbolic.dof_desc.UNNAMED_SOURCE` and
+            :class:`~pytential.symbolic.dof_desc.UNNAMED_TARGET` for
```
inducer, line 146: ...unspecified values...

pytential/symbolic/dof_desc.py
```diff
@@ -23,6 +23,9 @@
 from typing import Any, Hashable, Optional, Union
 
 __doc__ = """
+.. autoclass:: UNNAMED_SOURCE
+.. autoclass:: UNNAMED_TARGET
+
```
inducer, line 28:
```suggestion
```

pytential/symbolic/dof_desc.py
```diff
@@ -55,12 +58,26 @@
 
 # {{{ discretizations
 
+class UNNAMED_SOURCE:                   # noqa: N801
```
inducer, line 61: Link to the use case in the interface for `prepare_...`. (same for `_UNNAMED_TARGET`)

inducer (approved):
LGTM subject to the changes above.

pytential/symbolic/execution.py
```diff
@@ -37,6 +37,7 @@
 from pytential.symbolic.compiler import Code, Statement, Assign, ComputePotential
 
 from pytential import sym
+from pytential.symbolic.dof_desc import UNNAMED_SOURCE, UNNAMED_TARGET
 
 import logging
```
inducer, line 42: - Delete `domains` arg on `build_matrix`
- Delete `domains` arg on `scipy_op`.

alexfikl: Why? Weren't those supposed to support multiple geometries with different properties? (which, yeah, nobody uses or tests at all.. but still!)

inducer: Hmmm, upon closer reflection, I think I may have jumped the gun on removing `domains`. I think it has a role to play still. What tripped me up is that there should be `source_domains` and `target_domains`, technically. Where it says `DEFAULT_TARGET` must be a key, instead say something like "target" must be the result of an unnamed binding, such as might result from passing a tuple to `bind`.

IOW, @alexfikl, you're completely right.

alexfikl (approved):

inducer:
Thx!

PR #201, opened 2023-05-01
Make temporaries in OperatorCompiler

isuruf:
So that the scheduler can manage the life-time of the dependencies better.

For eg:
```
a = IntG0
b = IntG1
c = IntG2
res = a + b + c
```
is written as
```
a = IntG0
b = IntG1
temp = a + b
c = IntG2
res = temp +c
```
First plan needs all of `a, b, c` to be alive at the same moment, but the second plan does not.


inducer:
(Important bit of context from @isuruf : Naive Stokes made 36 temporaries before doing anything with them, at substantial, unnecessary memory cost)

isuruf:
@inducer, this is ready

inducer:
Thanks!

PR #203, opened 2023-05-05
Move pyfmmlib to an extra_requires

alexfikl:
I ran the tests locally without `pyfmmlib` and everything seems to be working, i.e. they're skipped and the nothing else crashes.


inducer:
Thx!

PR #208, opened 2023-05-30
Switch sumpy branch to main

alexfikl:
@isuruf Was this on purpose?


inducer:
That was entirely an accident, and I missed that.

inducer:
Thanks for pointing it out.

PR #210, opened 2023-06-19
skeletonization: fix crash with low number of proxies

alexfikl:
Finally got to hunting down and fixing the crash we were getting with a small number of proxies!

Basically, managed to hit a case where the `src_mat` had shape `(21, 23)` and the `tgt_mat` had shape `(23, 22)`. The rank after skeletonizing the target was `22`, which was smaller then the max one for the source and caused `interp_decomp` to crash somewhere in the Fortran code.

### DIFF
--- a/pytential/linalg/skeletonization.py
+++ b/pytential/linalg/skeletonization.py
@@ -616,6 +616,7 @@ def _skeletonize_block_by_proxy_with_mats(
         k = id_rank
         src_mat = np.vstack(src_result[i])
         tgt_mat = np.hstack(tgt_result[i])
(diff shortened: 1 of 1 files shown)


jezdez:
This should be good to go as pluggy has been yanked from conda-forge again: https://github.com/conda-forge/admin-requests/pull/737

alexfikl:
> This should be good to go as pluggy has been yanked from conda-forge again: [conda-forge/admin-requests#737](https://github.com/conda-forge/admin-requests/pull/737)

That sure was fast! Thanks a lot!

inducer:
Thank you!

PR #212, opened 2023-06-23
Reduce deprecation warnings

alexfikl:
https://github.com/inducer/pytential/actions/runs/5314367888/jobs/9621723155 had 1701079 warnings, which seems excessive, so this tries to reduce them a bit.

### DIFF
--- a/pytential/linalg/skeletonization.py
+++ b/pytential/linalg/skeletonization.py
@@ -132,7 +132,9 @@ def _approximate_geometry_waa_magnitude(
             <> ioffset = starts[icluster]
             <> npoints = starts[icluster + 1] - ioffset
             result[icluster] = reduce(sum, i, waa[indices[i + ioffset]]) / npoints
-            """)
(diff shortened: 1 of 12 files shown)


inducer:
LGTM, thanks!

alexfikl:
> LGTM, thanks!

This gets it down to about 13000 warnings, which seems a bit more manageable!

inducer:
I mean, that's still a crazy number... but :tada: progress!

alexfikl:
> I mean, that's still a crazy number... but tada progress!

Another ~2k should go away once inducer/pyopencl#689 gets into a release, but the rest are from `TaggableCLArray`, so those will have to wait a while :(

inducer:
https://github.com/inducer/pyopencl/releases/tag/v2023.1.1. Should be on pypi once https://github.com/inducer/pyopencl/actions/runs/5371394328 completes.

PR #219, opened 2023-07-31
Add mypy checking

alexfikl:
Takes over #123 and finishes up the mypy checking. This seems to work, but most places are of course still not type annotated :(


inducer:
Also added a Gitlab job: https://gitlab.tiker.net/inducer/pytential/-/merge_requests/new?merge_request%5Bsource_branch%5D=towards-mypy

inducer:
Thanks for getting us started down this road!

PR #223, opened 2023-09-09
tests: disable high memory test case

alexfikl:
This is running with the `pymbolic@attrs` branch to see if there's any test case we can disable to get that running on the Github CI (seems to have memory issues). Ran
```
mprof run python -m pytest -v -s -m 'not slowtest' --memray --most-allocations=25 test
```
on `koelsch` to see which cases use the most memory. There's a bunch of tests at around 2GB a piece, so going through them to see if it works :(

TODO:
- [ ] Revert the pymbolic branch change in `requirements.txt`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,8 +85,9 @@ jobs:
         -   uses: actions/checkout@v3
         -   name: "Main Script"
             run: |
+                export PYTEST_PARALLEL_FLAGS='-n 2'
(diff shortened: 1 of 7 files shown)


alexfikl:
Hm, well this isn't great. I disabled the top 6 (this far) biggest memory tests (according to `pytest-memray`) and it's still failing..

inducer:
Thanks for digging! A kind of desperate, but maybe effective way might be to bisect by file...?

PR #228, opened 2024-04-09
Do not install a conda env for flake8

inducer:


alexfikl:
And remove the `action/setup-python` where it doesn't make sense? :grin:

inducer:
Good point, done.

PR #230, opened 2024-07-04
Switch to pyproject and ruff

alexfikl:
This does the switch as for the other projects. I've tested that it still builds the sdist and wheel correctly with `python -m build`.

The only other thing that's changed is the removal of that `_git_rev.py` file. If I understood the logic correctly:
* In a normal git editable install, it was being overwritten anyway.
* In a tagged release sdist/wheel, it wouldn't matter because the full version is enough of an identifier.
* In a non-release sdist/wheel, it would still be needed, since there's no git directory. Is this a valid usecase? Maybe we can use something standardized like [setuptools-scm](https://github.com/pypa/setuptools_scm/) for that..

TODOs
* [x] Gitlab CI started at https://gitlab.tiker.net/fikl2/pytential/-/pipelines/563826.
* [x] mypy failures are fixed by https://github.com/inducer/pytools/pull/240. Specifically by changing `memoize_in(Generic[P, R])` to just `memoize_in`. Not sure why that's suddenly an issue..
	- Nevermind, apparently this is also fixed by setting `python_version = "3.8"` for mypy (like it was originally in `setup.cfg`). By default it uses the version of the interpreter (3.12 on the CI, I assume) and was unhappy about this for some reason..


alexfikl:
Gitlab CI started at https://gitlab.tiker.net/fikl2/pytential/-/pipelines/563826.

pyproject.toml
```diff
@@ -1,3 +1,171 @@
...
+
+[tool.ruff.lint]
+extend-select = [
+    "B",    # flake8-bugbear
+    "C",    # flake8-comprehensions
+    # "D",    # pydocstyle
+    "E",    # pycodestyle
+    "F",    # pyflakes
+    "G",    # flake8-logging-format
+    # "I",    # flake8-isort
```
alexfikl, line 89: The isort fixes are not added because that would likely cause a lot of conflicts in the opened PRs :(

pytential/version.py
```diff
@@ -20,28 +20,12 @@
...
-if os.environ.get("AKPYTHON_EXEC_FROM_WITHIN_WITHIN_SETUP_PY") is not None:
-    # We're just being exec'd by setup.py. We can't import anything.
-    _git_rev = None
+from pytools import find_module_git_revision
 
-else:
-    import pytential._git_rev as _git_rev_mod
-    _git_rev = _git_rev_mod.GIT_REVISION
+VERSION_TEXT = metadata.version("pytential")
+VERSION = tuple([int(i) for i in VERSION_TEXT.split(".")])
```
alexfikl, line 28: I noticed after I added this that the other projects have some regex for the version. Should this follow?

setup.py
```diff
@@ -83,61 +25,9 @@ def write_git_revision(package_name):
...
-        "sumpy>=2020.2beta1",
-
-        "scipy>=1.2.0",
-        "immutables",
-    ],
-    extras_require={
-        "fmmlib": ["pyfmmlib>=2019.1.1"],
-    },
-)
+setup(ext_modules=cythonize(ext_modules))
```
alexfikl, line 33: Apparently there's no way to cythonize from `pyproject.toml` :cry:

alexfikl:
@inducer Just re-ran the whole CI and it didn't crash, so not sure what that symengine failure was about :\

pytential/version.py
inducer: This is OK, too.

inducer:
Thx!

PR #231, opened 2024-07-04
ci: use github action concurrency to cancel jobs

alexfikl:
Mostly inspired from
https://docs.github.com/en/actions/examples/using-concurrency-expressions-and-a-test-matrix
Basically should help with me pushing and rebasing in a quick succession that just generates a lot of jobs on the CI :(

### DIFF
--- a/pytential/muller.py
+++ b/pytential/muller.py
@@ -55,7 +55,7 @@ def muller_deflate(
     # Truncates the zero arrays created above if necessary.
     for i in range(n):
         miter = 0
-        roots0, niter0 = muller(f_deflated, maxiter=maxiter, tol=eps)
+        roots0, niter0 = muller(f_deflated, maxiter=maxiter, tol=eps, z_start=z_start)
         roots.append(roots0)
         niter.append(niter0)
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,10 @@ on:
     schedule:
         - cron:  '17 3 * * 0'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
     ruff:
         name: Ruff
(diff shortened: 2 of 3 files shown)


alexfikl:
@inducer Ok, something is wrong with that muller test :(
In #230, the only thing that changed there was the random number generator, so not sure what would annoy it :\ I'll try to run it locally to see if it ever triggers a failure.

alexfikl:
@inducer Ran that test a few times and found a value that fails consistently in [e13bc79](https://github.com/inducer/pytential/pull/231/commits/e13bc79265e122d1c1ca9e32d1ad42ad5faa1305). I haven't looked into what's wrong there though..

pytential/muller.py
```diff
@@ -55,7 +55,7 @@ def f_deflated(z: T) -> T:
     # Truncates the zero arrays created above if necessary.
     for i in range(n):
         miter = 0
-        roots0, niter0 = muller(f_deflated, maxiter=maxiter, tol=eps)
+        roots0, niter0 = muller(f_deflated, maxiter=maxiter, tol=eps, z_start=z_start)
```
alexfikl, line 58: For what it's worth, this will also remove the failures on the CI. Basically
* Before, we used `np.random.seed` to set the seed globally so this call always initialized `z_start` to the same random value.
* Now with the change to `np.random.Generator`, this call without `z_start` was essentially unseeded, so it would sometimes hit a value that failed.

alexfikl: Maybe this is enough? From a nice blast from the past:
https://gitlab.tiker.net/inducer/pytential/-/issues/26

inducer: Yeah, let's just pick a good seed and move on.

alexfikl: Done in [70c7ed8](https://github.com/inducer/pytential/pull/231/commits/70c7ed842ddeed03dab90aba2c613419481ac72f).

inducer: Thanks!

PR #233, opened 2024-07-19
Add typos CI and fix issues

alexfikl:
Mostly found some typos looking through #162 and thought it might be a good idea to add it here too.


inducer:
Thx!

PR #234, opened 2024-07-20
beltrami: fix mypy partial issues

alexfikl:
mypy 1.11 added some changes around typing `functools.partial` that seems to be more strict about what arguments to pass in there. For `sym.S` and friends, I think they were confused by the `**kwargs` potentially providing a density too, not sure..

https://github.com/python/mypy/blob/master/CHANGELOG.md#support-for-functoolspartial


inducer:
Thx!

PR #237, opened 2024-08-11
Fix new C409 errors

alexfikl:
This fixes the failure from ruff 0.5.7 with rule C409 and fixes two extra rules in there, since they were straightforward.


inducer:
Thx!

PR #240, opened 2024-09-24
Remove deprecated CLMathFunction

alexfikl:
This removes the `CLMathFunction` expression and updates some docs to default to the numpy names of trigonometric functions (these were slated for removal in 2022 :D).


inducer:
Thx!

PR #245, opened 2024-11-15
Add matrix builders to docs

alexfikl:
Started using these in some of the direct solver annotations, so thought it's a good time to add them to the docs.

### DIFF
--- a/pytential/symbolic/primitives.py
+++ b/pytential/symbolic/primitives.py
@@ -558,7 +558,8 @@ class NumReferenceDerivative(DiscretizationProperty):
     def __new__(cls,
                 ref_axes: int | tuple[tuple[int, int], ...] | None = None,
                 operand: Operand | None = None,
-                dofdesc: DOFDescriptor | None = None) -> "NumReferenceDerivative":
+                dofdesc: DOFDescriptor | None = None,
+                ) -> "NumReferenceDerivative":
         # If the constructor is handed a multivector object, return an
         # object array of the operator applied to each of the
         # coefficients in the multivector.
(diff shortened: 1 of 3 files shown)


pytential/symbolic/primitives.py
```diff
@@ -567,7 +568,9 @@ def __new__(cls,
             def make_op(operand_i):
                 return cls(ref_axes, operand_i, as_dofdesc(dofdesc))
 
-            return componentwise(make_op, operand)
+            # FIXME: mypy is right: new should return `cls` instances and we're
+            # abusing it to vectorize the call like this.
+            return componentwise(make_op, operand)  # type: ignore[return-value]
```
alexfikl, line 573: Not quite sure what else to do about this?

inducer: https://docs.python.org/3/reference/datamodel.html#object.__new__ says:

> If [`__new__()`](https://docs.python.org/3/reference/datamodel.html#object.__new__) does not return an instance of cls, then the new instance’s [`__init__()`](https://docs.python.org/3/reference/datamodel.html#object.__init__) method will not be invoked.

I read this as implicit permission for `__new__` to return whatever it wants.

What happens if you're honest in the type annotation for `__new__`?

alexfikl: At least mypy is unhappy
```
pytential/symbolic/primitives.py:558: error: "__new__" must return a class instance (got "NumReferenceDerivative | ndarray[Any, dtype[Any]]")  [misc]
```

inducer: Mypy issue/PR rabbit hole in reverse chronological order:

- https://github.com/python/mypy/issues/15182
- https://github.com/python/mypy/pull/7188
- https://github.com/python/mypy/issues/1020

Including a BDFL quote:

> Honestly I don't think that A() should ever return something that's not an instance of A (or a subclass), even though technically __new__() can return anything it likes.

I don't think there's any way that mypy will learn what's going on here in the short term. Stylistically, it may be better to move away from overloading `__new__` and maybe instead have the user call `componentwise` directly. But that's a different discussion, #246.

alexfikl: He does agree that this sort of symbolic simplification usage is quite pleasant though :grin:
https://github.com/python/mypy/issues/1020#issuecomment-1081401853

PR #247, opened 2024-11-16
Replace __new__ componentwise

### LINKED ISSUE #246: Discussion: Deprecate overloading `__new__` for operator application?
Suppose `x` is a numpy object array of densities.

Now: `S(x)` returns a vector of single-layer potentials.

A possible future: `componentwise(S, x)` returns a vector of single-layer potentials.

Context/motivation: https://github.com/inducer/pytential/pull/245#discussion_r1844481562

cc @alexfikl

alexfikl:
Fixes #246.

### DIFF
--- a/pytential/symbolic/primitives.py
+++ b/pytential/symbolic/primitives.py
@@ -20,11 +20,11 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from collections.abc import Iterable
+from collections.abc import Callable, Iterable
 from dataclasses import field
 from warnings import warn
 from functools import partial
-from typing import Any, Literal, Union
+from typing import Any, Concatenate, Literal, TypeAlias, TypeVar
 
 import numpy as np
 
@@ -38,7 +38,9 @@ from pymbolic.geometric_algebra import MultiVector, componentwise
 from pymbolic.geometric_algebra.primitives import (
         NablaComponent, Derivative as DerivativeBase)
 from pymbolic.primitives import make_sym_vector
+from pymbolic.typing import ArithmeticExpressionT
 
+from pytools import P
 from pytools.obj_array import make_obj_array, flat_obj_array
 from sumpy.kernel import Kernel, SpatialConstant
 
@@ -94,6 +96,12 @@ associated with a :class:`~meshmode.discretization.Discretization`, then
     :show-inheritance:
     :undoc-members:
     :members: mapper_method
+
+.. autofunction:: for_each_expression
+
+.. class:: P
+
+    See :class:`pytools.P`
 
 Diagnostics
 ^^^^^^^^^^^
@@ -191,29 +199,43 @@ Elementary numerics
     :undoc-members:
     :members: mapper_method
 
+.. autofunction:: num_reference_derivative
+
 .. autoclass:: NodeSum
     :undoc-members:
     :members: mapper_method
+
+.. autofunction:: node_sum
 
 .. autoclass:: NodeMax
     :undoc-members:
     :members: mapper_method
 
+.. autofunction:: node_max
+
 .. autoclass:: NodeMin
     :undoc-members:
     :members: mapper_method
+
+.. autofunction:: node_min
 
 .. autoclass:: ElementwiseSum
     :undoc-members:
     :members: mapper_method
 
+.. autofunction:: elementwise_sum
+
 .. autoclass:: ElementwiseMin
     :undoc-members:
     :members: mapper_method
 
+.. autofunction:: elementwise_min
+
 .. autoclass:: ElementwiseMax
     :undoc-members:
     :members: mapper_method
+
+.. autofunction:: elementwise_max
 
 .. autofunction:: integral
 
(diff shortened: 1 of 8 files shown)


inducer:
Thanks! One thing that went through my head is whether we should outright deprecate direct constructor calls that bypass the factory functions.

inducer:
Implied but not said: how do you feel about that idea?

alexfikl:
> Thanks! One thing that went through my head is whether we should outright deprecate direct constructor calls that bypass the factory functions.

Calling the constructor directly should still be ok for "internal" code though, right? e.g. all the mappers call the constructor directly.

pytential/symbolic/primitives.py
```diff
@@ -1166,9 +1224,26 @@ def __post_init__(self) -> None:
...
+def interpolate(operand: ExpressionT,
+                from_dd: DOFDescriptorLike,
+                to_dd: DOFDescriptorLike) -> ExpressionT | Interpolation:
+    from_dd = as_dofdesc(from_dd)
+    to_dd = as_dofdesc(to_dd)
+
+    if from_dd == to_dd:
+        return operand
+
+    return Interpolation(from_dd, to_dd, operand)
```
alexfikl, line 1244: Is this ok? Introduced the new one just because the argument order didn't work :\

inducer:
> Calling the constructor directly should still be ok for "internal" code though, right? e.g. all the mappers call the constructor directly.

Yes, definitely. One question is whether we'll make them pass a flag to say, "I'm internal!"

pytential/symbolic/primitives.py
```diff
@@ -347,9 +372,26 @@
...
+    def make_stringifier(self, originating_stringifier=None):
+        from pytential.symbolic.mappers import StringifyMapper
+        return StringifyMapper()
+
+
+Operand: TypeAlias = (
+    ArithmeticExpressionT | np.ndarray[Any, np.dtype[Any]] | MultiVector)
 QBXForcedLimit = int | Literal["avg"] | None
 
+ExpressionT = TypeVar("ExpressionT", bound=ArithmeticExpressionT)
```
alexfikl, line 393: This needs a bit of thought? I'm not sure our naming scheme is sufficiently smart :grin:

inducer: I'm not sure I understand the issue. Could you explain?

alexfikl: I just meant that it's a typevar bound to `ArithmeticExpressionT` and it's called `ExpressionT` + it's different to the `ExpressionT` in pymbolic.

inducer: OIC. Good question... Addressing this properly would amount to establishing a new convention, which I'm kind of hesitant about. I guess we sort of had a convention of naming type variables `SomethingT`, but then I named `ExpressionT` the way I did... because `Expression` was taken. :thinking: PEP8 has [guidance](https://pep8.org/#type-variable-names) on naming type variables, but nothing that would help in this case. Do you have an opinion?

alexfikl: Yeah, I've also mostly gone with naming type variables as `SomethingT`, but besides that, can't say I have any useful ideas. 

Maybe `ExpressionLike` would have worked instead of `ExpressionT`? Since it could also potentially fit `MultiVector` and other things that mainly mappers know how to handle. `ExpressionNode` to emphasize the dag part?

alexfikl: Is renaming `ExpressionT` and/or `ArithmeticExpressionT` on the table? It's definitely way easier to just rename this one variable and not break the world :))

inducer: We'd have to probably leave it in place deprecated for a bit, because the type names are referred to in `cast`s. I'm not sure I love `ExpressionLike`. Maybe `ExpressionOrData` for `ExpressionT`?  And just `ArithmeticExpression` for `ArithmeticExpressionT`?

alexfikl: > I'm not sure I love ExpressionLike. Maybe ExpressionOrData for ExpressionT?

Can't say I have a big preference there.. I'll try to keep it in the back of my head and see if anything comes to mind :\

> And just ArithmeticExpression for ArithmeticExpressionT?

Yep!

In the meantime, I've renamed the one here to not be completely confusing.

inducer: For posterity: this discussion became https://github.com/inducer/pymbolic/pull/162.

pytential/symbolic/primitives.py
inducer: Hmm. I generally prefer having the "how" parameters before the "what" ones. At the same time, there is precedent for "what-before-how" (`ElementwiseMin`), so what do I know? It seems that this at least beats some consistency into things, so I'm OK.

inducer:
This is awesome. Thanks so much for doing this!

pytential/symbolic/primitives.py
alexfikl: @inducer Yeah, I remembered we did the `interp` one in that order on purpose. 

Some ways to make that work would be to 
* make `for_each_expression` work like a multimap, so that it would vectorize over all arrays in `*args`.
* make `for_each_expression(vec_args=n)` to be able to tell it which argument is meant to be vectorized.

Does any of that sound good?

inducer: I'm not sure any of those could be statically typed.

alexfikl: Hm, yeah, fair point, can't think of a way to do that.

For comparison, this is what `jax.jit` is doing (where e.g. `static_argnums` works in a similar way to my second point):
https://github.com/jax-ml/jax/blob/b372ce4b1ab0bee7a1da495b098ff3948a6c0d4d/jax/_src/api.py#L141=L155

inducer: The new `interpolate` can be typed straightforwardly, so I don't see a good reason to deviate from it.

PR #248, opened 2024-11-18
Improve compat with pre-dataclass expressions

alexfikl:
Mostly adds back some `dofdesc=None` default constructors found during merging main into #162.

### DIFF
--- a/pytential/symbolic/mappers.py
+++ b/pytential/symbolic/mappers.py
@@ -61,7 +61,7 @@ def rec_int_g_arguments(mapper, expr):
             name: mapper.rec(arg) for name, arg in expr.kernel_arguments.items()
             }
 
-    changed = (
+    changed = not (
             all(d is orig for d, orig in zip(densities, expr.densities, strict=True))
             and all(
                 arg is orig for arg, orig in zip(
--- a/pytential/symbolic/primitives.py
+++ b/pytential/symbolic/primitives.py
@@ -20,10 +20,11 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
+from collections.abc import Iterable
 from dataclasses import field
 from warnings import warn
 from functools import partial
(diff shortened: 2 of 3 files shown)


pytential/symbolic/mappers.py
```diff
@@ -61,7 +61,7 @@ def rec_int_g_arguments(mapper, expr):
             name: mapper.rec(arg) for name, arg in expr.kernel_arguments.items()
             }
 
-    changed = (
+    changed = not (
```
alexfikl, line 64: I do not know how this didn't break the world before.. :facepalm:

pytential/symbolic/mappers.py
```diff
@@ -280,7 +280,12 @@ def map_common_subexpression(self, expr):
 # {{{ FlattenMapper
 
 class FlattenMapper(FlattenMapperBase, IdentityMapper):
-    pass
+    def map_int_g(self, expr):
+        densities, kernel_arguments, changed = rec_int_g_arguments(self, expr)
+        if not changed:
+            return expr
+
+        return replace(expr, densities=densities, kernel_arguments=kernel_arguments)
```
alexfikl, line 288: Also from #162, it would be nice to teach `ExpressionKernel` in sumpy to flatten a bit more. Not sure what the best place for that is..

inducer, line 288: Why is this needed? It looks an awful lot like what would be inherited from `IdentityMapper`? (If there is a reason, please add a comment.)

alexfikl: Yeah, it is. It didn't work due to https://github.com/inducer/pytential/pull/248#discussion_r1847138517 and I forgot to revert this part.

Removed in https://github.com/inducer/pytential/compare/165d0e140a2538ededcbcf1545b4c21205f27e34..baba46fa9af2222a69b1816438274a443e1baa32

pytential/symbolic/mappers.py
inducer: Agreed. Scary... Thanks for finding it!

PR #250, opened 2024-12-05
ruff: disable rule RUF052

alexfikl:
RUF052 seems to complain about underscore variables in function arguments. We use that quite a bit to denote "useful, but private, don't depend on this" arguments, so it seemed like a good idea to disable it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,6 +98,7 @@ extend-ignore = [
     "N806", # variable name should be lowercase
     "RUF012", # ClassVar for mutable class attributes
     "RUF022", # __all__ is not sorted
+    "RUF052", # underscored variable is used
     "UP031", # use f-strings instead of %
     "UP032", # use f-strings instead of .format
 ]


alexfikl:
The pylint failure is quite exciting. No idea what that's about..

From what I can tell, it's due to `3.x` finally getting bumped to `3.13` :tada:

inducer:
I'm not super happy with RUF052 as it is: https://github.com/astral-sh/ruff/issues/14796. Let's see what happens.

alexfikl:
> I'm not super happy with RUF052 as it is: [astral-sh/ruff#14796](https://github.com/astral-sh/ruff/issues/14796). Let's see what happens.

Yeah, agreed. It feels very weird to apply that rule to function/method arguments.

inducer:
The pylint failure seems to be this: https://github.com/pylint-dev/pylint/issues/10000#issuecomment-2519899671

alexfikl:
I'll go ahead and close this, since hopefully it'll just fix itself in the next week or so :grin:

PR #251, opened 2024-12-08
Fix some more deprecations from primitives

alexfikl:
The ruff and pylint failures will still be there :(

### DIFF
--- a/pytential/qbx/target_specific/impl.pyx
+++ b/pytential/qbx/target_specific/impl.pyx
@@ -129,7 +129,7 @@ def h3dall_wrapper(nterms, z, scale, hs, hders):
 
 # {{{ helpers
 
-cdef void legvals(double x, int n, double[] vals, double[] derivs) nogil:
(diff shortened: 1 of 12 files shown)


pytential/qbx/target_specific/impl.pyx
```diff
@@ -129,7 +129,7 @@ def h3dall_wrapper(nterms, z, scale, hs, hders):
 
 # {{{ helpers
 
-cdef void legvals(double x, int n, double[] vals, double[] derivs) nogil:
+cdef void legvals(double x, int n, double[] vals, double[] derivs) noexcept nogil:
```
alexfikl, line 132: No idea if adding `noexcept` to these is reasonable. From a quick skim, it doesn't seem like anything in there would raise an exception.. :\

inducer: I think that's fine. I'd like to remove that code before too long anyway, so if it still passes tests, that's good enough for me.

inducer:
Thanks!

PR #252, opened 2024-12-15
Fix RUF052 errors

alexfikl:
These should also be fixed by https://github.com/astral-sh/ruff/issues/14968, but it's not a big change to update them here.


inducer:
Thanks!

PR #254, opened 2025-01-06
Bump tolerance for skeletonization test

alexfikl:
This should fix the new failures. As far as I can tell, this is due to scipy 1.15 having a rewritten `interp_decomp` implementation.
```
>               assert tgt_error < ctol * case.id_eps
E               AssertionError: assert np.float64(7.672159794924751e-08) < (6 * 1e-08)
E                +  where 1e-08 = TorusTestCase(name='torus', ambient_dim=3, knl_class_or_helmholtz_k=0, knl_kwargs={}, bc_type='dirichlet', side=-1, qb...ighted_proxy=None, proxy_source_cluster_builder=None, proxy_target_cluster_builder=None, neighbor_cluster_builder=None).id_eps
```

### DIFF
--- a/pytential/linalg/utils.py
+++ b/pytential/linalg/utils.py
@@ -326,7 +326,10 @@ def make_flat_cluster_diag(
 # {{{ interpolative decomposition
 
 def interp_decomp(
-        A: np.ndarray, *, rank: int | None, eps: float | None,
+        A: np.ndarray, *,
+        rank: int | None,
+        eps: float | None,
+        rng: np.random.Generator | None = None,
         ) -> tuple[int, np.ndarray, np.ndarray]:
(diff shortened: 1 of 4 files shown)


pytential/linalg/utils.py
```diff
@@ -337,11 +340,21 @@ def interp_decomp(
...
         raise ValueError("providing both 'rank' and 'eps' is not supported")
 
+    from scipy import __version__
+
+    kwargs = {}
+    if __version__ >= "1.15.0":
+        if rng is None:
+            rng = np.random.default_rng()
+
+        kwargs["rng"] = rng
```
alexfikl, line 350: Had to start using `rng` in here because `scipy.linalg.interpolative.seed` is a noop on the latest version, so the tests kept failing due to randomness.

inducer:
LGTM, thanks!

PR #255, opened 2025-01-14
quadrature comparison

ShawnL00:


alexfikl (commented):
This is very cool! Just ran it locally to see the errors and there's a pretty big mismatch :\

Left some comments in there, but there's no big issues, just nitpicks. I imagine `ruff` is going to complain about some things, so might be worth running it locally until the CI recovers.

ShawnL00 (commented):
Thank you for the detailed feedback, Alex. I have run `ruff`, and it seems fine now.

alexfikl:
> Thank you for the detailed feedback, Alex. I have run `ruff`, and it seems fine now.

That's great! Looks good to me now :grin:

inducer:
It seems that jemalloc was at least involved in the problem. I've [disabled it](https://github.com/inducer/ci-support/commit/b01966bbe69b0bbb291e72428ae140623033dece).

examples/planewave-quad-comparison.py
```diff
@@ -0,0 +1,197 @@
...
+
+            # Tensor quadrature
+            qbx_order = tensor_target_order
+            sources_h, weights_nodes_h = quadrature(level,
+                            tensor_target_order, qbx_order=qbx_order,
+                            group_cls=TensorProductElementGroup)
+            tensor_value = np.dot(wave(sources_h), weights_nodes_h)
+            tensor_result.append(tensor_value)
+            tensor_nodes.append(len(sources_h[0]))
+            tensor_err.append(np.abs(tensor_value - ref))
```
inducer, line 165: This should be a loop (since it's doing the same thing twice).

examples/planewave-quad-comparison.py
```diff
@@ -0,0 +1,197 @@
...
+      Exp[I r (n[[1]] Sin[θ] Cos[φ] +
+               n[[2]] Sin[θ] Sin[φ] +
+               n[[3]] Cos[θ])];
+
+    NIntegrate[
+     wave[θ, φ] * r^2 Sin[θ],
+     {θ, 0, Pi},
+     {φ, 0, 2 Pi},
+     WorkingPrecision -> 21
+    ]
```
inducer, line 73: Could you try to avoid using Mathematica? The main problem is that I can't reproduce the result, or tweak the computation in any way.

https://mpmath.org/doc/current/calculus/integration.html#mpmath.quad should work just as well.

examples/planewave-quad-comparison.py
```diff
@@ -0,0 +1,197 @@
...
+from arraycontext import flatten
+from meshmode.array_context import PyOpenCLArrayContext
+import pyopencl as cl
+
+cl_ctx = cl.create_some_context()
+queue = cl.CommandQueue(cl_ctx)
+actx = PyOpenCLArrayContext(queue)
+
+
+def quadrature(level, target_order, qbx_order, group_cls=SimplexElementGroup):
```
inducer, line 97:
```suggestion
def quadrature(level, target_order, qbx_order, group_cls):
```

examples/planewave-quad-comparison.py
```diff
@@ -0,0 +1,197 @@
...
+10        17             21
+11        19             23
+12        20             25
+13        22             27
+14        24             29
+15        25             31
+16        27             33
+17        28             35
+18        30             37
+19        32             39
```
inducer, line 35: It's not useful to incorporate this table, since it's not guaranteed to match the code on an ongoing basis. Instead, the code should generate the table.

examples/planewave-quad-comparison.py
inducer, line 97: The way this is currently expressed is flawed: At no point do you explicitly specify which quadrature should be used. And what pytential uses internally is an implementation detail that could change at any time. If you're intending to measure differences between specific quadrature rules, your code should exert positive control over what rule is used, instead of relying on implementation coincidence. It may be easier to do this by using meshmode primitives.

examples/planewave-quad-comparison.py
```diff
@@ -0,0 +1,197 @@
...
+def quadrature(level, target_order, qbx_order, group_cls=SimplexElementGroup):
+    mesh = mgen.generate_sphere(1, target_order,
+            uniform_refinement_rounds=level, group_cls=group_cls)
+    pre_density_discr = Discretization(actx, mesh,
+                    InterpolatoryQuadratureGroupFactory(target_order))
+
+    qbx = QBXLayerPotentialSource(
+        pre_density_discr, target_order, qbx_order,
+        fmm_order=False)
+    discr_stage = sym.QBX_SOURCE_STAGE1
```
inducer, line 106: Using pytential's stage-1 is a potential confounder, since it may apply refinement to the incoming discretization.

examples/planewave-quad-comparison.py
```diff
@@ -0,0 +1,197 @@
...
+    places = GeometryCollection({"qbx": qbx}, auto_where="qbx")
+    density_discr = places.get_discretization("qbx", discr_stage)
+    ambient_dim = qbx.ambient_dim
+    dofdesc = sym.DOFDescriptor("qbx", discr_stage)
+
+    sources = density_discr.nodes()
+    weights_nodes = bind(places,
+            sym.weights_and_area_elements(ambient_dim=3, dim=2, dofdesc=dofdesc))(actx)
```
inducer, line 114: Please document how you convinced yourself that the error is not dominated by geometry derivatives.

inducer (commented):
Thanks! Some comments below.

PR #259, opened 2025-01-31
Use constantdict instead of immutables

alexfikl:
Very small usecase here, but might as well switch over.


pytential/collection.py
```diff
@@ -170,47 +170,47 @@ def __init__(self,
             assert isinstance(places, Mapping)
             places_dict = places
 
-        self.places = immutables.Map(places_dict)
-        self.auto_where = (auto_source, auto_target)
-
-        self._caches: dict[str, Any] = {}
-
```
alexfikl, line 177: This is just a bit of churn that moves all the `self.attr = value` after the validation was done.

PR #260, opened 2025-02-16
Fix missing keyword arguments in generate_torus

alexfikl:
Broke this in https://github.com/inducer/meshmode/pull/450 :(


inducer:
Thx!

PR #273, opened 2025-08-29
Build docs in CI

alexfikl:
Was missing the `maybe_upload_docs` like boxtree.


alexfikl:
The `flat_obj_array` tries to fix the failure from
https://gitlab.tiker.net/inducer/pytential/-/pipelines/706241

inducer:
Thx!

PR #274, opened 2025-08-30
Add more types to primitives

alexfikl:
Some remaining issues (for another time):
* From `MultiVector` not having all the methods typed (`map` and stuff)
* From `Derivative` not being typed (all that `dnabla` machinery).
* Typing these made a lot of errors in other places, so they might not be quite right. Although the stats on `baseline.json` show a lot more errors are removed, so probably not that big of an issue.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -276,14 +276,6 @@
             {
                 "code": "reportUnknownParameterType",
                 "range": {
-                    "startColumn": 4,
-                    "endColumn": 16,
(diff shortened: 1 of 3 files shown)


inducer:
Heh! I'd started down a similar path as a result of the #273 failures, but only got as far as https://github.com/inducer/pytools/pull/320 plus some scattered changes on my local machine. I've pushed them to #275. If you like, feel free to grab them and incorporate them here.

inducer:
Thx!

PR #275, opened 2025-08-30
A few types in primitives

inducer:


alexfikl:
This was added to #274.

PR #242, opened 2024-10-08
Port to new pymbolic

alexfikl:
Updates things to use `expr_dataclass` and fixes various deprecation warnings.

Needs:
- [X] https://github.com/inducer/pymbolic/pull/150
- [X] https://github.com/inducer/sumpy/pull/217
- [X] https://github.com/inducer/loopy/pull/868

PR #280, opened 2025-09-27
Improve typing in symbolic.pde.scalar

alexfikl:


pytential/symbolic/primitives.py
```diff
@@ -1681,6 +1690,25 @@ def area(ambient_dim: int,
...
+def mean(ambient_dim: int,
+         dim: int,
+         operand: ObjectArrayND[ArithmeticExpression],
+         dofdesc: DOFDescriptorLike = None) -> ObjectArrayND[ArithmeticExpression]: ...
+
+@overload
+def mean(ambient_dim: int,
+         dim: int,
+         operand: MultiVector[ArithmeticExpression],
+         dofdesc: DOFDescriptorLike = None) -> MultiVector[ArithmeticExpression]: ...
```
alexfikl, line 1709: Is there a way to do this without adding a bunch of overloads?

pytential/symbolic/pde/scalar.py
```diff
@@ -190,53 +264,61 @@ def __init__(self, kernel: Kernel, loc_sign: int, *,
...
-        :param u: symbolic variable for the density.
-        :param kwargs: additional keyword arguments passed on to the layer
-            potential constructor.
-        """
-        sqrt_w = self.get_sqrt_weight()
+            map_potentials = default_map_potentials
+
+        kernel_arguments = self.kernel_arguments
+        if kwargs:
+            kernel_arguments = {**kernel_arguments, **kwargs}
```
alexfikl, line 298: Before, `kwargs` and `kernel_arguments` would get passed down to `IntG`, which would raise a `ValueError` if some keys repeated. From a quick search, it doesn't seem like anyone is using those extra `kwargs` anyway, but is it ok to let it overwrite?

alexfikl: Hm, actually `int_g_vec` already does exactly this since #242 (also my bad!). So I guess these things got overwritten for a while now..

pytential/symbolic/primitives.py
inducer: Update: does exist.

PR #281, opened 2025-10-03
Improve typing in pytential.source and pytential.compiler

alexfikl:


pytential/symbolic/compiler.py
```diff
@@ -24,85 +24,125 @@
...
+
+    .. automethod:: get_assignees
+    .. automethod:: get_dependencies
     """
 
-    names: list[str]
-    exprs: list[Expression]
+    names: tuple[str, ...]
+    """Names of the variables used in this statement."""
+    exprs: tuple[Expression, ...]
```
alexfikl, line 81: Made a few of these to requiring tuples, but hopefully it doesn't break anything (the classes themselves are already frozen).

pytential/source.py
```diff
@@ -76,51 +87,32 @@ def ndofs(self) -> int:
...
-        return expr
-
-
-class _SumpyP2PMixin:
-
-    def get_p2p(self,
-                actx: PyOpenCLArrayContext,
-                target_kernels: tuple[Kernel, ...],
-                source_kernels: tuple[Kernel, ...] | None = None) -> P2P:
-        @memoize_in(actx, (_SumpyP2PMixin, "p2p"))
```
alexfikl, line 123: What was the purpose of this `SumpyP2PMixin`? I tried to do some git blaming, but couldn't find any explanation.

Removed because it displeased `pyright` :grin:

alexfikl:
@inducer This should pass now (it was just an overzealous `assert` that failed before), so you can take a look when you have the time.

pytential/source.py
inducer: Very unclear. I looked through the history, and I don't get it either.

pytential/symbolic/compiler.py
```diff
@@ -24,85 +24,125 @@
...
+.. autoclass:: Statement
+.. autoclass:: Assign
+.. autoclass:: PotentialOutput
+.. autoclass:: ComputePotential
+
+.. autofunction:: dot_dataflow_graph
+
+.. autoclass:: Code
+.. autoclass:: OperatorCompiler
+"""
```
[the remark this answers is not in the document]
inducer: It's OK to document them IMO.

pytential/qbx/__init__.py
inducer, line 1: These changes should not be in a commit named "update baseline".

pytential/symbolic/compiler.py
```diff
@@ -340,22 +362,28 @@ def gen_expr_arrow(expr, target_node):
...
-            inputs: Set[str],
+            inputs: set[str],
             schedule: Sequence[tuple[Statement, Collection[str]]],
-            result: np.ndarray,
+            result: Expression | ObjectArray1D[Expression],
            ) -> None:
         self.inputs = inputs
-        self._schedule = schedule
         self.result = result
+        self._schedule = schedule
```
inducer: Looks like this guy could be a `dataclass`.

inducer: I would like this to have a type parameter, with bound `Expression | ObjectArray1D[Expression]`, so that we can still tell whether it's a scalar or a vector output. This will help with typing precision of operator outputs.

inducer (commented):
Thanks!

pytential/qbx/__init__.py
alexfikl: :facepalm: Fixed in https://github.com/inducer/pytential/compare/e298c4c35806a9263d579fd906207a1ca0864956..f2e479958d61bfce633afd450ef21e87295cec7f.

pytential/symbolic/compiler.py
inducer: Fair. Though I'm happy to type for *intent* and fix the code if it doesn't actually do what the intended types suggest.

PR #320, opened 2026-06-02
Use OneKernel directly

alexfikl:
This replaced `one_kernel_2d` and `one_kernel_3d` (not actually used) with direct calls to `OneKernel`. These seem to be the only places where those `one_kernel_Nd` things are used (including `sumpy`).


inducer:
Thx!

PR #322, opened 2026-06-21
Early interp (density)

ShawnL00:
WIP:

This PR adds need-driven interpolation for layer-potential densities.

Plain densities continue to be interpolated as whole units from stage1 to quad_stage2. For densities coupled with geometry factors, this PR adds explicit markers:

 - `sym.distribute_interpolation(expr)` allows interpolation to be pushed into selected density expressions.
 - `sym.geo_density(geometry, density)` is a helper for common `geometry * unknown` densities.
 - `sym.interpolation_unit(expr)` marks quantities that must remain atomic under interpolation.

For example, this lets expressions such as `normal * sigma` be written so that the normal is formed using the stage2 geometry parameterization and interpolated to quad_stage2, while the unknown `sigma` is interpolated from stage1 to quad_stage2. At the same time, QWeight/L2-scaled quantities such as `u / sqrt_jac_q_weight` remain atomic.

`tangential_to_xyz(...)` transforms variables from a local tangential basis to the standard Cartesian basis, so the resulting Cartesian vector should be interpolated as an atomic quantity.

### DIFF
--- a/pytential/symbolic/primitives.py
+++ b/pytential/symbolic/primitives.py
@@ -51,10 +51,11 @@ from pymbolic.geometric_algebra.primitives import (
     NablaComponent,
 )
 from pymbolic.primitives import (  # noqa: N813
+    CommonSubexpression,
     Variable as var,
     cse_scope as cse_scope_base,
     expr_dataclass,
-    make_common_subexpression as cse,
+    make_common_subexpression as _cse,
     make_sym_vector,
 )
 from pymbolic.typing import ArithmeticExpression
@@ -92,7 +93,7 @@ if TYPE_CHECKING:
 
     import modepy as mp
     from pymbolic.mapper.stringifier import StringifyMapper
-    from pymbolic.primitives import CommonSubexpression, Quotient
+    from pymbolic.primitives import Quotient
 
 
 __doc__ = """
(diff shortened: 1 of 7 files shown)


pytential/symbolic/primitives.py
```diff
@@ -2498,7 +2583,8 @@ def tangential_to_xyz(
     tonb = tangential_onb(ambient_dim, dofdesc=dofdesc)
     result = sum(tonb[:, i] * tangential_vec[i] for i in range(ambient_dim - 1))
 
-    return cast("ObjectArray1D[ArithmeticExpression]", result)
+    return cast("ObjectArray1D[ArithmeticExpression]",
+            interpolation_unit(result, "tangential_to_xyz"))
```
inducer: - Not sure we need `InterpolationUnit`?
- In this particular instance, it seems like using it and not using it will be equivalent (given that all the geo information is polynomials anyway), and in a hypothetical world with access to exact geometry, it would make things worse?

pytential/symbolic/primitives.py
```diff
@@ -527,6 +545,73 @@ class ErrorExpression(ExpressionNode):
     """The error message to raise when this expression is encountered."""
 
 
+@expr_dataclass()
+class InterpolationUnit(CommonSubexpression):
```
inducer:
```suggestion
class InterpolationUnit(ExpressionNode):
```

pytential/symbolic/primitives.py
ShawnL00: You're right. I will remove it. The coefficients (and the discretized geometry) are polynomials.

inducer:
The conda issue might be resolved (without my having done anything).

PR #174, opened 2022-08-04
Direct Solver: Multilevel Proxy-based Solver

alexfikl:
Attempting to break up the direct solver MR into smaller, more reviewable pieces. As a rundown

* **[Done]** Add proxy generation (already in ``linalg.proxy``).
* **[Done]** Add single-level skeletonization (#30).
* **[WIP]** Add multi-level skeletonization (#164).
* **[WIP]** Add the actual multi-level direct solver (#174).

The code for this is mostly in the deprecated https://gitlab.tiker.net/inducer/pytential/-/merge_requests/137.

PR #325, opened 2026-06-23
Typing 2026 06 23

inducer:


alexfikl:
This seems to fix the macOS failure for me: [`faa5a1e` (#174)](https://github.com/inducer/pytential/pull/174/changes/faa5a1e40fcf93cb013b78cf7e0f04ca8807afd8).

Found some similar looking errors in https://github.com/prefix-dev/pixi/issues/3479 that seem to suggest that bumping the compiler to 14.2.0 also works, so maybe https://github.com/inducer/ci-support/pull/42 also does the trick :shrug:

inducer:
Thanks for taking a look! Let's see how the mac CI does with the ci-support change merged.

alexfikl:
> Thanks for taking a look! Let's see how the mac CI does with the ci-support change merged.

There's still a `+++ conda install --yes ld64=609` line in there, so I don't think it trickled down yet?

inducer:
Thanks for spotting that! The gitlab push had failed due to some admin thing I'd done yesterday. Should be all better now, retrying.

PR #327, opened 2026-06-27
Port Stokes to new SystemKernels

alexfikl:
Probably better viewed commit by commit, since the second commit adds type annotations and is a bit noisy.


.basedpyright/baseline.json
```diff
@@ -39960,2968 +39960,6 @@
                 }
             }
         ],
-        "./pytential/symbolic/stokes.py": [
```
alexfikl, line 39867: 🎉

inducer:
Thanks for making this! Do you think it would also be possible to replace all the `*Wrapper` classes with a generic system-kernel applier? If not, what are the obstacles?

alexfikl:
Well, using the Stokes kernels as an example (and the component order from Pozrikidis), we'd need 
```math
\begin{aligned}
\mathcal{S} & = \int G_{ij}(x, y) q_j(y), \\
\mathcal{S}^* & = \int G_{ji}(x, y) q_j(y)
\end{aligned}
\qquad 
\begin{aligned}
\mathcal{D} & = \int n_k(y) T_{ijk}(x, y) q_i(y), \\
\mathcal{D}^* & = \int n_k(x) T_{ijk}(x, y) q_j(y) 
\end{aligned}
```
where the star ones are just the adjoints. I haven't looked too much how these look for $`\mathcal{S}', \mathcal{S}'', \mathcal{D}'`$ since they're not very used, but once these are constructed we can probably just extend things like `sym.normal_derivative` to work on them (?).

Just thinking out loud a bit.. ideally, we'd have something like
```python
apply(system_kernel, density, direction1, direction2, direction3, ...)
```
We can probably get away with letting the `SystemKernel` return a tuple of `ndim - 1` indices to dot into with the first one being the density and the rest being the direction vectors. This should allow using 
```python
u = apply(stokeslet, density)
traction = apply(stresslet, density, normal, adjoint=True)
D = apply(stresslet, density, normal, adjoint=False)
sigma = apply(stresslet, density, adjoint=True)
```
to compute the double layer potential or the traction or the full stress tensor (i.e. skipping the normal just doesn't dot). Something like that? Not sure what else would be needed here..

inducer:
So, in order to make the `apply` generic, given your example, we'd need something like `density_contraction_axis` (axes?) and `normal_contraction_axis` (left-over would implicitly be "output")?

alexfikl:
> So, in order to make the `apply` generic, given your example, we'd need something like `density_contraction_axis` (axes?) and `normal_contraction_axis` (left-over would implicitly be "output")?

Yeah, separating them is fine. The normal here would be specifically for the double layer / normal derivative type usage, right?

One thing that I'm not sure about is some of my older Stokes work, where I wanted to get the tangential stress, which, if I recall correctly was something like this
```math
\tau = t_k(x) \int T_{ijk}(x, y) q_j(y) dy
```
Is this worth shoving in here somehow or is it too specific? i.e. I can contract with the density to get the "stress tensor" and then it's my job (not pytential's) to know how to contract to get the normal/tangential stress.

test/test_symbolic.py
```diff
@@ -603,6 +603,56 @@ def test_derivative_with_spatial_constant():
 # }}}
 
 
+# {{{ test_system_kernel_int_g
+
+@pytest.mark.parametrize("dim", [2, 3])
+def test_system_kernel_int_g(dim: int) -> None:
```
inducer: Unresolved interface issues:

- How would one do a double-layer (e.g. Laplace)?
- `stress` case doesn't work.
- More of Isuru's use cases.

PR #330, opened 2026-07-26
Fix new implicit-string-concatenation-in-collection-literal

alexfikl:
This also adds a little test for `dot_dataflow_graph` because I wasn't sure if those two lines needed a semicolon or not :grin:


inducer:
Thx!

PR #331, opened 2026-08-07
Fix small issues in primitives.py

alexfikl:
Probably best viewed commit by commit. It just moves things around and updates some phrasing/typos.

The baseline changes are due to moving some things around + probably from the `boxtree` FMM typing PR. (`updated ./.basedpyright/baseline.json with 5430 errors (went down by 82)` :grin:)


pytential/symbolic/primitives.py
```diff
@@ -2031,22 +2062,19 @@ def __post_init__(self) -> None:
...
-        if missing_args:
+        if missing_args := (kernel_arg_names - provided_arg_names):
             raise ValueError(
                 "Kernel argument(s) not supplied: '{}'".format(", ".join(missing_args))
             )
 
-        # FIXME: this check is clearly wrong :(
-        extra_args = provided_arg_names - kernel_arg_names
-        if missing_args:
+        if extra_args := (provided_arg_names - kernel_arg_names):
```
alexfikl, line 2070: I forgot to fix this a long time ago. It's correct now, right?

alexfikl: Hm, this fails some tests currently. The issue is that e.g. `NeumannOperator` for Helmholtz takes in some `kernel_arguments` for the kernel, but the representation also uses a `LaplaceKernel`, which is going to complain because it got an extra `k` when it gets passed the same `kernel_arguments`.

I would be inclined to convert the raise into a warning? I don't know..

I've currently just removed the `kernel_arguments` from that call (in [a4295c9](https://github.com/inducer/pytential/pull/331/commits/a4295c91941c2bea9a82719ec3a925d13a66dd4b)), so it should work :shrug:

inducer: IMO, extra arguments *should* be an error, so this is headed in the right direction for my liking.

inducer:
Thanks!

PR #333, opened 2026-08-19
Make flatten work on Operands

alexfikl:
It seemed like some places where using `componentwise(flatten, expr)`, but this looks nicer + one less import :grin:


inducer:
Thx!

